PR #4, opened 2021-05-05
reintroduced missing conditional to sts template. fixes #3

### LINKED ISSUE #3: Chart does not work on its own due to hard-coded postgres url
in the sonarqube-sts.yaml template, there is a busybox job that tries to connect to your postgresql server using a hardcoded value:

```
command: ["/bin/sh", "-c", "for i in $(seq 1 200); do nc -z -w3 {{ .Release.Name}}-postgresql 5432 && exit 0 || sleep 2; done; exit 1"]
```
this causes the pod to never start.  I had to change this to

```
command: ["/bin/sh", "-c", "for i in $(seq 1 200); do nc -z -w3 {{ .Values.postgresql.postgresqlServer}} 5432 && exit 0 || sleep 2; done; exit 1"]
```

to make it work.   hardcoding the "-postgresql" in the name is a bad idea.  I'm assuming this is due to the optional bundled postgresql server option, but I'm using an external server.

tobias-trabelsi-sonarsource:


pierre-guillot-sonarsource (commented):
Hello @tobias-trabelsi-sonarsource , Do you thing it would make sense to track this work/change in a Jira ticket?

pierre-guillot-sonarsource (approved):

PR #5, opened 2021-05-05
Manage case where prometheus is disbaled

fredrikand:
With prometheus disabled, prometheus Config was still added to volumes list.
This caused POD to never be able to start since the Config for prometheus did not exist.


charts/sonarqube/templates/sonarqube-sts.yaml
```diff
@@ -111,6 +111,10 @@ spec:
             - mountPath: /data
               name: sonarqube
               subPath: data
+          {{- with .Values.env }}
```
tobias-trabelsi-sonarsource, line 114: any reason you need to have the general env here and not something specific to the scope of the init container? i think it would make more sense to create a separate scope for this if this is even necessary

tobias-trabelsi-sonarsource (changes requested):
Thank you very much for your contribution and the fix for the bad mount. 
i do have a remark regarding the env introduction for the inject-prometheus-exporter init container thou

charts/sonarqube/templates/sonarqube-sts.yaml
fredrikand: Agree. Something more specific would be better. Will update

When running in environments where the only path to internet is through proxy, some way to specify the proxy settings is needed to get the prometheus init container to complete.

fredrikand: Realized that the  mountpoint  /opt/sonarqube/conf was actually used twice when activating prometheus export. No go :-)
Moved prometheus.yaml to sonarqube config instead.

Also, busybox wget does not support https proxy. Using curl instead.

tobias-trabelsi-sonarsource:
please revert the changes to the jmx exporter config location and the configmap to create said config. see the CI for error details and i could confirm in my local cluster that these changes result in a crash of the init container as the filesystem in /data is not present at the execution time of the init container.
the changes to switch from wget to curl seem fine for me. just please add them to the config summary in the `README.md` file

tobias-trabelsi-sonarsource (changes requested):
please revert the changes to the jmx exporter config location and the configmap to create said config. see the CI for error details and i could confirm in my local cluster that these changes result in a crash of the init container as the filesystem in /data is not present at the execution time of the init container.
the changes to switch from wget to curl seem fine for me. just please add them to the config summary in the `README.md` file

tobias-trabelsi-sonarsource:
Hi @fredrikand ,
i addressed your points in this PR here: #10 
thanks for pointing them out and the initial work on fixing them

PR #10, opened 2021-05-14
Prometheus init container

tobias-trabelsi-sonarsource:
fixed Prometheus config volume mount if disabled
switched from wget to curl image per default for downloading agent
added support for proxy envs

PR #16, opened 2021-05-15
change to use curl image for plugin download

### LINKED ISSUE #15: [Q] outdated alpine-wget image
I have noticed that you are using the `rjkernick/alpine-wget` image, which is quite outdated and does not seem to be maintained anymore.
Due to this I spun up a [very slim repo myself](https://github.com/k3rnelpan1c-dev/alpine-wget) with as much automation as possible to keep it up to date, that is based on a current version of Alpine that I am using for our install of this chart. (due to security limitations placed upon the images we can deploy)

If you want to use a up to date image feel free to use mine or copy the behaviour found in #10 and switch to the curl images entirely. (note: you could have also kept the wget there as long as you would have used the `rjkernick/alpine-wget`, which thanks to the full install of wget respects the proxy settings)

k3rnelpan1c-dev:
### Description

Since the currently used `rjkernick/alpine-wget:latest` is unmaintained it can be seen as a security risk and might get blocked by security restrictions in certain deployments. To circumvent this this PR replicates what #10 started and moved the only use of the image to use a up to date `curl` image. This in turn required an update of the plugin install script to use `curl`.

### Changes

* migrates plugin-install init container to use `curlimages/curl:7.76.1` instead of `rjkernick/alpine-wget:latest`
* updated plugin install script to use `curl` instead of `wget`
* change how proxy settings are handled for the plugin script (unifies it with the Prometheus init container) 
* fix a possible bug with the Prometheus init container if `env` would have been set in values

### Issue(s)

* closes #15 

### Notes

* The `Deployment` kind is actually missing the Prometheus init container, is that intended?
* Proposal: unify the 2 proxy settings in the `values.yaml` since its unlike to have 2 different proxies set for each of them

### DIFF
--- a/charts/sonarqube/templates/install-plugins.yaml
+++ b/charts/sonarqube/templates/install-plugins.yaml
@@ -9,21 +9,12 @@ metadata:
     heritage: {{ .Release.Service }}
 data:
   install_plugins.sh: |-
-    {{- if .Values.plugins.httpProxy }}
-    export http_proxy={{ .Values.plugins.httpProxy }}
-    {{- end }}
-    {{- if .Values.plugins.httpsProxy }}
-    export https_proxy={{ .Values.plugins.httpsProxy }}
-    {{- end }}
-    {{- if .Values.plugins.noProxy }}
-    export no_proxy={{ .Values.plugins.noProxy }}
-    {{- end }}
     {{- if .Values.plugins.install }}
       [ -e {{ .Values.sonarqubeFolder }}/extensions/downloads/* ] && rm {{ .Values.sonarqubeFolder }}/extensions/downloads/*
-      {{ range $index, $val := .Values.plugins.install }}
-      echo {{ $val | quote }} >> {{ $.Values.sonarqubeFolder }}/extensions/downloads/list{{ end }}
-      cat {{ .Values.sonarqubeFolder }}/extensions/downloads/list | xargs -n 1 -P 8 wget --directory-prefix {{ .Values.sonarqubeFolder }}/extensions/downloads --no-verbose{{ if .Values.plugins.noCheckCertificate }} --no-check-certificate{{ end }}
-      rm {{ .Values.sonarqubeFolder }}/extensions/downloads/list
+      cd {{ .Values.sonarqubeFolder }}/extensions/downloads
+      {{- range $index, $val := .Values.plugins.install }}
+      curl {{ if $.Values.plugins.noCheckCertificate }}--insecure{{ end }} -fsSLO {{ $val | quote }}
+      {{- end }}
     {{- end }}
     {{- if .Values.plugins.lib }}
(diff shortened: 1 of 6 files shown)


tobias-trabelsi-sonarsource:
> The Deployment kind is actually missing the Prometheus init container, is that intended?

time is everything :sweat_smile: i plan to unify both of these but currently i am lacking the time to do so.

i will take a closer look at your PR when i am back at the office, but it looks fine so far :+1:

k3rnelpan1c-dev:
I totally feel you on the time bit :sweat_smile:.

charts/sonarqube/templates/install-plugins.yaml
```diff
@@ -9,21 +9,12 @@ metadata:
...
-    {{- end }}
     {{- if .Values.plugins.install }}
       [ -e {{ .Values.sonarqubeFolder }}/extensions/downloads/* ] && rm {{ .Values.sonarqubeFolder }}/extensions/downloads/*
-      {{ range $index, $val := .Values.plugins.install }}
-      echo {{ $val | quote }} >> {{ $.Values.sonarqubeFolder }}/extensions/downloads/list{{ end }}
-      cat {{ .Values.sonarqubeFolder }}/extensions/downloads/list | xargs -n 1 -P 8 wget --directory-prefix {{ .Values.sonarqubeFolder }}/extensions/downloads --no-verbose{{ if .Values.plugins.noCheckCertificate }} --no-check-certificate{{ end }}
-      rm {{ .Values.sonarqubeFolder }}/extensions/downloads/list
+      cd {{ .Values.sonarqubeFolder }}/extensions/downloads
+      {{- range $index, $val := .Values.plugins.install }}
+      curl {{ if $.Values.plugins.noCheckCertificate }}--insecure{{ end }} -fsSLO {{ $val | quote }}
```
tobias-trabelsi-sonarsource, line 16: i know that this has previously also been downloaded without checking certificates, but now reading it again i think it might make sense to leave the option, but now be insecure by default.
can you take care about that in this PR? else i will handle this separated

k3rnelpan1c-dev: I am a bit confused as to what you are requesting, do you want me to:
* change it so it always uses `--insecure`
* checking if this now is defaulting to `--insecure` (it should not right now unless `.Values.plugins.noCheckCertificate` is set to true)
* my opinion on defaulting too `--insecure`, which I am no fan of due to security concerns NGL
* update documentation

I am open for updating it whatever way you want me to, just require some clarification :wink:

tobias-trabelsi-sonarsource: oh sorry i missed that if statement :see_no_evil: this was exactly what i wanted.

tobias-trabelsi-sonarsource (approved):

PR #18, opened 2021-05-18
prometheusExporter.downloadURL (custom download URL for the agent jar)

thomasgl-orange:
This PR adds a new config value for explicitly providing the `jmx_prometheus_javaagent.jar` download URL.
When this (optional) value is set, it overrides the default download URL from Maven Central (meaning the `prometheusExporter.version` value is ignored).

Motivations:
- allow deployment without direct internet access, using only some internal mirror / proxy repo
- allow using a patched / SNAPSHOT exporter agent version


tobias-trabelsi-sonarsource (changes requested):
Thanks for the implementation. the changes in general do make sense, but could you add the modifications to the Changelog and bump the version number?

thomasgl-orange:
> could you add the modifications to the Changelog and bump the version number?

Sure, done (f4c81fceb81c39bc6df7f3b173a9a5b4f2c891d5). Thanks for having a look at this PR.

PR #19, opened 2021-05-21
Update with nginx-ingress config and 1.17 compatibility

ja6a-regnosys:
* update ingress for Kubernetes server 1.17 compatibility
* remove nginx dependency
* include basic values.yaml with:
  * nginx-ingress configuration
  * standard sonarqube config


ja6a-regnosys:
Needs more work

PR #32, opened 2021-06-09
fix: minor issues

### LINKED ISSUE #30: too long container name
Hi,

You need to fix container name about password-hook.

```
$ helm install -n sonarqube-stage sonarqube-stage-sonarqube -f sonarqube-stage-sonarqube-values.yaml ../helm-chart-sonarqube/charts/sonarqube/
Error: failed post-install: warning: Hook post-install sonarqube/templates/change-admin-password-hook.yml failed: Job.batch "sonarqube-stage-sonarqube-sonarqube-change-admin-password-hook" is invalid: spec.template.spec.containers[0].name: Invalid value: "sonarqube-stage-sonarqube-sonarqube-change-default-admin-password": must be no more than 63 characters
```

### LINKED ISSUE #31: sts.yaml does not have imagePullSecrets
Hi,

You need to add below code in sts.yaml

I can find this in deployment but  not in sts.
```
{{- if .Values.image.pullSecret }}
      imagePullSecrets:
        - name: {{ .Values.image.pullSecret }}
{{- end }}
```

k3rnelpan1c-dev:
### Description

This PR aims to fix 2 of the reported Issues with the this chart.
It also artificially limits the full sonarqube deployment name to 33 characters as a proposed fix for #30, however this can also be solved differently and I am open for feedback.

### Changes

* add missing imagePullSecrets to sts
* fix too long names globally (on a `_helpers.tpl` level)

### Issues

* fixes #31

### DIFF
--- a/charts/sonarqube/CHANGELOG.md
+++ b/charts/sonarqube/CHANGELOG.md
@@ -1,8 +1,11 @@
 # SonarQube Chart Changelog
 All changes to this chart will be documented in this file.
 
+## [1.0.9]
+* add missing imagePullSecrets in sts install type
+
 ## [1.0.8]
-* fix typo in initfs 
+* fix typo in initfs
 * fix plugin installation init container permissions
 * fix duplicated mount point for conf when sonar.properties are defined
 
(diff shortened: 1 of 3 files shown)


tobias-trabelsi-sonarsource:
you're on fire @k3rnelpan1c-dev :D 

i am a bit hesitant regarding #30 . this seems to me like a user config error rather than something that needs to be taken care of in the helm chart. i don't have the feeling that the names are unreasonable long currently.
Also i think the `trunc 63` -> `trunc 33` will only reduce the probability of the error described but not fix it right? 

i am open for a discussion here

PR #37, opened 2021-06-16
Prometheus JMX exporter agent for CE process

thomasgl-orange:
Currently (with chart version `1.0.10`), enabling `prometheusExporter.enabled` only enables the [JMX exporter agent](https://github.com/prometheus/jmx_exporter) on the **Web** process. With this PR, it will be enabled on the **CE** process too. This second agent will, of course, listen on a different port (8001 instead of 8000).

By default, both agent will use an identical configuration (`prometheusExporter.config`). If required, an optional `prometheusExporter.ceConfig` value can be set to override the CE process agent configuration specifically.

**Is that useful?**

I think so. In addition to the usual JVM stuff, you'll get these metrics to monitor the CE activity:
```
SonarQube_ComputeEngineDatabaseConnection_PoolActiveConnections
SonarQube_ComputeEngineDatabaseConnection_PoolIdleConnections
SonarQube_ComputeEngineDatabaseConnection_PoolInitialSize
SonarQube_ComputeEngineDatabaseConnection_PoolMaxActiveConnections
SonarQube_ComputeEngineDatabaseConnection_PoolMaxIdleConnections
SonarQube_ComputeEngineDatabaseConnection_PoolMaxWaitMillis
SonarQube_ComputeEngineDatabaseConnection_PoolMinIdleConnections
SonarQube_ComputeEngineDatabaseConnection_PoolRemoveAbandoned
SonarQube_ComputeEngineDatabaseConnection_PoolRemoveAbandonedTimeoutSeconds
SonarQube_ComputeEngineTasks_ErrorCount
SonarQube_ComputeEngineTasks_InProgressCount
SonarQube_ComputeEngineTasks_PendingCount
SonarQube_ComputeEngineTasks_ProcessingTime
SonarQube_ComputeEngineTasks_SuccessCount
SonarQube_ComputeEngineTasks_WorkerCount
SonarQube_ComputeEngineTasks_WorkerMaxCount
```
(+ `SonarQube_ComputeEngineTasks_LongestTimePending` with the next `jmx_exporter` version; see prometheus/jmx_exporter#587)

**Rational for the configuration choices**
- the agent is enabled by default for Web, so enabling by default for CE seems coherent
- this could be done via an other independent option (`prometheusExporter.ceEnabled`, or something like that), but then users who have explicitly set `prometheusExporter.enabled: false` with previous versions of the chart, and don't pay attention to changelogs, would suddenly get the default behavior (enabled) for the CE agent. Could be annoying (especially in environments where the agent can't be downloaded at all).
- my initial idea for configuration was actually something like that:
```yaml
prometheusExporter:
  # global settings with default values, for both Web and CE
  enabled: true
  config:
    rules:
      - pattern: ".*"
  # allows overriding settings for the Web and/or CE process
  #webEnabled: true
  #webConfig: ...
  #ceEnabled: true
  #ceConfig: ...
```
But it turned out it was too much boilerplate for little benefits (maybe because I'm not fluent at writing Helm, but also because helm/helm#3308).

Hence the simpler option implemented here, with only one new value:
- you can enable either both agents, or none
- you can, but don't have too, set up a specific config for the CE agent

**Why not enabling the agent for ES process too, while at it?**
Because it doesn't work. (A side effect of the Security Manager used in ElasticSearch, I guess). To monitor the ES process, an option would be to add a pod with the [ElasticSearch Exporter](https://github.com/prometheus-community/elasticsearch_exporter), but it's out of scope for this PR.


tobias-trabelsi-sonarsource (approved):
Hi @thomasgl-orange ,

thanks for this great PR. everything makes sense to me. 
regarding your comment on you prometheus JMX repository PR; i forwarded this to one of our backend engineers. in the future feel free to create a report on our community forum for triage for such cases

PR #38, opened 2021-06-17
restart SQ when only modifying the exporter agents config

thomasgl-orange:
When deploying a change for the Prometheus exporter agents (just `prometheusExporter.config` and/or `prometheusExporter.esConfig`, and nothing else), SQ doesn't restart.  Too bad I did not see that before submitting #37...

Fixed by adding more checksum annotations.


tobias-trabelsi-sonarsource (approved):
Good catch!

PR #42, opened 2021-06-24
add a Prometheus exporter for ElasticSearch

thomasgl-orange:
Following #37 which added the JMX exporter agent to the CE process, here is what I propose for adding an ElasticSearch exporter. It runs [elasticsearch_exporter](https://github.com/prometheus-community/elasticsearch_exporter) in an additional container of the `sonarqube` pod, using (by default) the [bitnami/elasticsearch-exporter](https://github.com/bitnami/bitnami-docker-elasticsearch-exporter) image.

Regarding configuration of the chart, this PR adds the following values:

```yaml
prometheusExporter:
  esMetricsPort: 8002
  elasticsearch:
    image: bitnami/elasticsearch-exporter:1.2.0
    esUri: http://localhost:9001
    extraArgs:
      - "--es.indices"
      - "--es.shards"
    securityContext:
      readOnlyRootFilesystem: true
      runAsNonRoot: true
    resources: ...
```

The ES exporter will be enabled/disabled together with the two other ones, via `prometheusExporter.enabled`.
I also wanted to keep the listen port value next to the other exporter ports (`prometheusExporter.{webBeanPort,ceBeanPort}`), hence the `prometheusExporter.esMetricsPort`. 
For other values (configuration of the container itself, and `elasticsearch_exporter` parameters), I thought it would be confusing to mix them with pre-existing `prometheusExporter.*` values (even if prefixed with `es`), hence the new `prometheusExporter.elasticsearch` subtree.  I'm not 100%sure it's the right choice, you let me know what you think.

Regarding backward compatibility, I think adding this container is okay. Even when it doesn't work as expected out-of-box (if the ElasticSearch address has been customized in `sonar.properties`), it should be fine, the exporter will still be running.

### DIFF
--- a/charts/sonarqube/CHANGELOG.md
+++ b/charts/sonarqube/CHANGELOG.md
@@ -1,6 +1,9 @@
 # SonarQube Chart Changelog
 All changes to this chart will be documented in this file.
 
+## [1.0.15]
+* added a Prometheus exporter for the ElasticSearch process
+
 ## [1.0.14]
 * added hostAliases to deploymentType statefulset
 
--- a/charts/sonarqube/Chart.yaml
+++ b/charts/sonarqube/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: sonarqube
 description: SonarQube is an open sourced code quality scanning tool
-version: 1.0.14
+version: 1.0.15
 appVersion: 8.9-community
 keywords:
   - coverage
--- a/charts/sonarqube/README.md
+++ b/charts/sonarqube/README.md
@@ -72,7 +72,17 @@ ingress:
 
 ## Monitoring
 
-This Helm chart offers the possibilitie to monitor SonarQube with Prometheus. Per default the jmx metrics for the Web Bean and the CE Bean are exposed on port 8000 and 8001. These Values can be configures with `prometheusExporter.webBeanPort` and `prometheusExporter.ceBeanPort`.
+This Helm chart offers the possibilitie to monitor SonarQube with Prometheus, with three exporters:
+
+- one for the *Web Server* process: a [jmx_exporter](https://github.com/prometheus/jmx_exporter), running as a JVM agent, and listening by default on port `8000` (`monitoring-web`), with metrics on `/`
+- one for the *Compute Engine* process: a [jmx_exporter](https://github.com/prometheus/jmx_exporter), running as a JVM agent, and listening by default on port `8001` (`monitoring-ce`), with metrics on `/`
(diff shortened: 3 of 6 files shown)


tobias-trabelsi-sonarsource:
Hi @thomasgl-orange ,

thanks for the PR. i had a look at it looks good to me. i am just a bit hesitant with the introduction of a sidecar container that exports generic ES metrics as these metrics do not seem to be that important as some are checked via `api/system/health`  and sonarqube will report to be unhealthy if something is wrong with ES anyway. Some other metrics, like `elasticsearch_filesystem_data_available_bytes` seem useful, but should probably better be monitored by a node exporter or directly in the PV.

Monitoring is a topic at the moment and i hope to get some momentum on that soon. i will discuss this topic with my PM on monday if it makes sense to introduce a generic ES exporter or if we can ship a better solution directly with SQ soon, so please stand by :)

tobias-trabelsi-sonarsource (commented):
Hi @thomasgl-orange ,

thanks for the PR. i had a look at it looks good to me. i am just a bit hesitant with the introduction of a sidecar container that exports generic ES metrics as these metrics do not seem to be that important as some are checked via `api/system/health` and sonarqube will report to be unhealthy if something is wrong with ES anyway. Some other metrics, like `elasticsearch_filesystem_data_available_bytes` seem useful, but should probably better be monitored by a node exporter or directly in the PV.

Monitoring is a topic at the moment and i hope to get some momentum on that soon. i will discuss this topic with my PM on Monday if it makes sense to introduce a generic ES exporter or if we can ship a better solution directly with SQ soon, so please stand by :)

thomasgl-orange:
> these metrics do not seem to be that important as some are checked via `api/system/health` and sonarqube will report to be unhealthy if something is wrong with ES anyway. Some other metrics, like `elasticsearch_filesystem_data_available_bytes` seem useful, but should probably better be monitored by a node exporter or directly in the PV.

I am mostly interested in the `elasticsearch_jvm_*` metrics actually, to check if our ES JVM sizing is okay, and detect any worrying trends as our SQ usage grows, before we reach the unhealthy point. I would have been happy simply using the JMX exporter, but it looks like the agent won't work on ES.

> Monitoring is a topic at the moment and i hope to get some momentum on that soon.

:+1: 

> i will discuss this topic with my PM on Monday if it makes sense to introduce a generic ES exporter or if we can ship a better solution directly with SQ soon, so please stand by :)

Sure, consider this PR as food for thought. I you end up adding a Prometheus-compatible endpoint directly in SQ, with metrics for the three JVMs (alike what's in `/api/system/info`), I will be more than happy too :)

tobias-trabelsi-sonarsource:
Hello @thomasgl-orange ,

i had a chat with my PM on this issue and we concluded that it would make more sense to solve this in a homogenized way in the future. There is [MMF-2022](https://jira.sonarsource.com/browse/MMF-2022) to keep an eye on for a solution on monitoring that is coming with SQ in the future.

thank you again for sending over this good PR :)

PR #43, opened 2021-07-05
JMX exporter 0.16.0

thomasgl-orange:
See https://github.com/prometheus/jmx_exporter/releases/tag/parent-0.16.0
This version includes prometheus/jmx_exporter#587, meaning it should enable exposing the `LongestTimePending` metric of the CE process.


tobias-trabelsi-sonarsource (commented):
Hi @thomasgl-orange ,

do you really think we need to change the default in the helm chart? i don't see the need to change this as it can be defined by the user anyway. Maybe you see something that i don't.

also please note that the optional was removed in 9.0 which was released on friday. i did not yet change the default application version and the image version because i am waiting on feedback for backports, but with that in mind i see even less value looking at the changelog

thomasgl-orange:
> Maybe you see something that i don't.

Nope, there's nothing more. The only purpose of this PR is to help ops of a 8.9.x LTS deployment get back the CE metric they are currently missing (probably without knowing).
If you don't think it's worth touching the chart, you can close the PR, it' okay :-)

tobias-trabelsi-sonarsource:
>The only purpose of this PR is to help ops of a 8.9.x LTS deployment get back the CE metric they are currently missing (probably without knowing).

This is a good point. can you rebase against master? i updated the kind action with the hope that this will fix the control plane to become ready in time for the QA

thomasgl-orange:
Sure, I've just rebased and pushed --force, let's see how it goes.

tobias-trabelsi-sonarsource (approved):
LGTM

PR #47, opened 2021-07-06
"Route" support for OpenShift deployments

gasgithub:
Added support for Route object to be used in case of OpenShift deployments instead of Ingress.

Changes:
- route.yamle in templates
- new section in values.yaml
- updated Notes to get the route URL

### DIFF
--- a/charts/sonarqube/templates/NOTES.txt
+++ b/charts/sonarqube/templates/NOTES.txt
@@ -3,6 +3,9 @@
 {{- range .Values.ingress.hosts }}
   http://{{ .name }}
 {{- end }}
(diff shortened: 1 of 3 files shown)


tobias-trabelsi-sonarsource (changes requested):
Hi @gasgithub ,

i think this is a nice addition, but please make sure that the QA is passing and that you don't overwrite the default values.
Regarding the required version bump i need to clarify in which version we can put this, as this was out of scope for the initial release

gasgithub:
Ok, thanks for comments. I'll revert defaults, bump minor for now, and create new PR.

PR #49, opened 2021-07-07
Route support for OpenShift deployments

gasgithub:
Added support for Route object to be used in case of OpenShift deployments instead of Ingress.

Changes:
- route.yaml in templates
- new section in values.yaml
- updated Notes to get the route URL

Based on previous comments:
Restored default settings in values.yaml. Set default to not create route. Bumped version to pass tests.


charts/sonarqube/values.yaml
```diff
@@ -86,6 +86,15 @@ ingress:
   #   hosts:
   #     - chart-example.local
 
+# Create Route for OpenShift deployments. Set enabled=true if you want Route to be automatically generated based on service and namespace names.
+route:
```
tobias-trabelsi-sonarsource, line 90: I think the new values should at least be mentioned in the `README.md`. don't you think?

tobias-trabelsi-sonarsource (changes requested):

charts/sonarqube/values.yaml
gasgithub: I added config params to README, although I think that most users will read the descriptions from the `values.yaml` file anyway.

tobias-trabelsi-sonarsource:
Hi @gasgithub ,

can you rebase and bump the chart version once more? after that i can merge your PR

charts/sonarqube/values.yaml
os-learner: I agree with you both. README.md is the main file anyone will read.

os-learner:
Thanks all.

PR #63, opened 2021-07-29
Adjust probes

jacek-poreda-sonarsource:
* increase initial delay
* increase failure threshold for startup probe

### DIFF
--- a/charts/sonarqube-dce/values.yaml
+++ b/charts/sonarqube-dce/values.yaml
@@ -118,13 +118,13 @@ ApplicationNodes:
   readinessProbe:
     initialDelaySeconds: 120
     periodSeconds: 30
-    failureThreshold: 6
+    failureThreshold: 8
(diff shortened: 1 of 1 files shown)


charts/sonarqube-dce/values.yaml
```diff
@@ -116,15 +116,15 @@ ApplicationNodes:
     runAsUser: 1000
 
   readinessProbe:
-    initialDelaySeconds: 120
+    initialDelaySeconds: 180
```
tobias-trabelsi-sonarsource, line 119: don't you think 3 minutes is a little much? this will block SQ from being available for 3 minutes regardless of that actual status. imo we should increase the `failureThreshold`

jacek-poreda-sonarsource: indeed, you are right 👍 I will adjust threshold here

tobias-trabelsi-sonarsource (approved):

PR #69, opened 2021-08-09
SONAR-15243 Enable json logging in k8s

jacek-poreda-sonarsource:


tobias-trabelsi-sonarsource (changes requested):
Just my 2 cents here: would it make sense to provide only one flag for the json output? i can not think of a reason why somebody would want to have a structured log for the search nodes but not for the application nodes.

in any case the new parameter needs to be added to the `README.md` of the chart

jacek-poreda-sonarsource:
Fair point, it could be one, I will add it to README as well 👍

charts/sonarqube-dce/values.yaml
```diff
@@ -93,6 +93,10 @@ ApplicationNodes:
     # Sonarqube dockerfile creates sonarqube user as UID and GID 1000
     runAsUser: 1000
 
+  # Enable/Disable logging in JSON format
+  logging:
```
tobias-trabelsi-sonarsource, line 97: i think this should go as well right? :slightly_smiling_face:

tobias-trabelsi-sonarsource (changes requested):

tobias-trabelsi-sonarsource (approved):
Perfect :slightly_smiling_face:

dbmeneses (approved):
nice :+1:

PR #76, opened 2021-08-19
BUILD-1107 polishing

tobias-trabelsi-sonarsource:
While working on our internal deployment i noticed two things: 

1. we can remove the `initSysctl` init container from the application pods
2. we can source the jwt secret for the application pods from an external secret

sadly this is not possible for the `monitoringPasscode`


jacek-poreda-sonarsource (approved):
Nice 👍

PR #74, opened 2021-08-11
Add ability to specify an optional array of containers to run alongside the `sonarqube` container

dlbock:
# Why

Please see https://github.com/SonarSource/helm-chart-sonarqube/issues/73 for the request and reasoning

# What

- Add an additional configuration parameter `extraContainers` to allow users to specify an array of containers to run alongside the `sonarqube` container either in the `StatefulSet` mode or `Deployment` mode.
- `charts/sonarqube/templates/deployment.yaml` contained duplicate definitions of `nodeSelector`, `tolerations` and `affinity`, therefore removed the duplicates.

PR #77, opened 2021-08-20
added route support. fixes #48

### LINKED ISSUE #48: Create "route" for OpenShift deployments
Would be nice to have option to create Route instead of Ingress fro OpenShift deployments :-)

Here is updated PR with this https://github.com/SonarSource/helm-chart-sonarqube/pull/49

tobias-trabelsi-sonarsource:
also added missing changelog from #74
supersedes #49


wouter-admiraal-sonarsource:
I have _no_ clue of what I'm looking at 🤷‍♂️ . So I don't see how I can review this. I'll let someone from the Asterix squad look into it, as they at least worked on the k8s topic.

k3rnelpan1c-dev (approved):
I may not have the power of a green check-mark, but this checks out as far as I can tell

jacek-poreda-sonarsource (approved):

PR #78, opened 2021-08-24
Update description of SonarQube

benzonico:
Include security scanning, because that's what we do !


tobias-trabelsi-sonarsource (changes requested):
Makes sense to me, but for sake of consistency, do you mind to also add this to the lts chart?

benzonico:
Sure thing ! Done !

tobias-trabelsi-sonarsource (approved):
LGTM.
No version bump needed as change is not functional

PR #89, opened 2021-09-09
Fix PVC `chart` label

sirspen:
**What happens:**
When attempting to apply this Helm Chart with `persistence.enabled` set to `true` I receive this error:

```
Error: PersistentVolumeClaim "sonarqube-sonarqube" is invalid: metadata.labels: Invalid value: "sonarqube-1.1.1+98": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')
```
This seems to be due to the label `chart` on the PVC, which is set to `{{ .Chart.Name }}-{{ .Chart.Version }}`. With the versioning currently on the chart, this adds a `+` to the label, which is not allowed in Kubernetes.

**What's expected:**
All resources deploy, including the PVC

**Fix:**
As with [other resources in this chart](https://github.com/SonarSource/helm-chart-sonarqube/blob/master/charts/sonarqube/templates/service.yaml#L7), replacing `+` with `_` should fix this issue.

### DIFF
--- a/charts/sonarqube-lts/CHANGELOG.md
+++ b/charts/sonarqube-lts/CHANGELOG.md
@@ -1,7 +1,10 @@
 # SonarQube Chart Changelog
 All changes to this chart will be documented in this file.
 
-## [1.0.16]
+## [1.1.3]
+* fixed `invalid: metadata.labels: Invalid value` error on the `chart` label of the pvc
+
+## [1.0.17]
 * release to helm repository
 * updated appversion to new LTS patch release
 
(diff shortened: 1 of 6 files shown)


tobias-trabelsi-sonarsource (changes requested):
Hi @sirspen ,

thanks for spotting the bug AND delivering a fix so quickly. would you mind bumping the chart version for this and mentioning your change in the `CHANGELOG.md`? 
I belive that the LTS might also be impacted by this. If you don't want to touch another chart, then i will take car of it :+1:

PR #94, opened 2021-09-15
Adding defaultBackend block to ingress spec

luis-ochoa:
The k8s Ingress spec v1 (https://kubernetes.io/docs/reference/kubernetes-api/service-resources/ingress-v1/#IngressSpec) has a provision to specify the default backend to be used when no host rules match the request. 

Some ingress classes can usually deduce this value if there is a single host rule in the spec with a standard port (80 or 443) but cannot make any assumptions when using non-standard ports (like sonarqube's 9000), resulting in warnings like:
Default backend:  default-http-backend:80 (<error: endpoints "default-http-backend" not found>)
or
Warning  PortResolutionError  69m (x6 over 3h53m)  azure/application-gateway  Code="ErrorUnableToResolveBackendPortFromServicePort" Message="No port matched ecommbgo/sonarqube-sonarqube" 

Granted, we can always make sonarqube service listen on port 80, but this PR gives the user the ability to add this block to the spec with the same service:port values as the chart's.


luis-ochoa:
While I agree that the _configuration driven_ block makes the chart  more complex, I also think that the chart shouldn't leave parts of the spec out of the user's reach. In my case I needed the block to quelch the warnings and I found no way of incorporating it.

Ideally the block should be generated automatically (just like the service block gets generated automatically). After all it won't hurt anyone; you're declaring one service and that's your default backend.
I can change the code to do that (no config, gets generated every time) or you can cancel the PR if you still don't agree.
Cheers,
Luis.

tobias-trabelsi-sonarsource:
i rather want to understand why you needed to have a default backend. for me this feels like a "catch all" when no rule was specified, but this charts specifies a rule for a backend, so i do not understand the need. 

yeah giving the user the option to do anything is nice, but there is a choice to make in between complexity and usability and in this case i do not see a clear point to increase complexity. if you can provide me a good reason i am all open for this change

luis-ochoa:
Agree with you. There's no need for the added complexity. In my case, I wanted to silence the dummy warnings from the ingress regarding the lack of default backend, and that's why I made the change. Thought of submitting it in case anybody else was in need of this.

tobias-trabelsi-sonarsource:
understood, and thank you for taking the time to prepare the PR. i will close this one then and make a note on our documentation for the next SQ release :slightly_smiling_face:

PR #95, opened 2021-09-15
Update for a better excerpt

ganncamp:
Didn't realize until I saw the published charts that I should have been working against a char limit for initial impact in the search results list


tobias-trabelsi-sonarsource (approved):
README changes don't need to pass CI. 

@ganncamp in the future you can indicate this with prefixing your commit with `[skip ci]`

PR #102, opened 2021-09-29
fix(postgres)!: ability to overwrite

jef:
## Description

Currently, when using `jdbcOverwrite.enable`, it doesn't change the behavior of PostgreSQL interaction. It still includes the PostgreSQL dependencies and spins up a PostgreSQL server. Along with using the default username and password even when `jdbcOverwrite.jdbcUsername` and `jdbcOverwrite.jdbcPassword` are specified.

Without having to flip a bunch of logic to check `jdbcOverwrite.enable` before `postgresql.enabled`, having `postgresql.enabled` set to `false` seems more reasonable, as it also gets rid of dependencies that aren't needed. (This is because `requirements.yaml` does not use advanced conditional logic).

Another interesting point is that when trying to override `postgresql.enabled` to `false`, this doesn't seem to work. Not sure if this is a Helm thing or a bug with the template.

Along with change:

- Various typo updates
- Formatting of table to match DCE
- Changed `jdbcOverwrite.enable` to `enabled` to match `postgres.enabled` and other variables that use `.enabled`


wouter-admiraal-sonarsource (changes requested):
Hi @jef, thank you for your contribution.

Before we review this, can you please make a few small changes?

1. You introduced quite a few whitespace changes in the Markdown files, and some of them actually break the formatting. Could you commit only the parts of the Markdown that relate to your fix, and leave any whitespace changes out, please?
2. The chart linting test fails, as it requires a version bump. Can you update this so the full test suite can run?

Thanks 🙏 . Once it is clean, we'll check your fix.

tobias-trabelsi-sonarsource (changes requested):
Hi @jef ,

i did have a closer look at your changes already and i think the formatting changes should be reverted in general. they are purely cosmetic and do not require a change.
apart from that the switch of the default of `postgresql.enabled` would break the out of the box deployment and should also be reverted.
as a last point you were changing `Values.jdbcOverwrite.enable` to `Values.jdbcOverwrite.enabled` without changing the default, so this is now undefined in the templates and also needs to be reverted. 

after that i do not see any other change that you made. am i missing something?

jef:
> You introduced quite a few whitespace changes in the Markdown files, and some of them actually break the formatting.

I can revert all the documentation change. Thanks!

> apart from that the switch of the default of `postgresql.enabled` would break the out of the box deployment and should also be reverted.

This is the main concern. I don't mind reverting the rest, but the problem here is that setting `postgresql.enabled: false` does not work. You can see for yourself by doing the following:

```shell
cd $(mktemp -d)
touch values.yaml
```

Feel free to test with either:

```yaml
postgres:
  enabled: false
```

or

```yaml
jdbcOverwrite:
  enabled: true
  jdbcUrl: jdbc:postgresql://testhost:5432/test
  jdbcUsername: sonarqube
  jdbcSecretName: sonarqube
  jdbcSecretPasswordKey: sonaqube-password
```

Then template out with the published Chart:

```shell
helm template --repo=https://SonarSource.github.io/helm-chart-sonarqube --version=1.1.6+121 -f ./values.yaml sonarqube
```

---

The output _does not_ disable Postgres. On top of that, it uses values from the default PostgreSQL configuration.

The main issue here is that either a) `jdbcOverwrite` checks need to happen first or b) `postgreql.enabled` needs to be `false`.

jef:
Made the changes, but not sure what version y'all would like to use. I decided to set to `2.0.0` since `postgresql.enabled` to `false` could be considered a breaking change.

tobias-trabelsi-sonarsource:
works for me:

`helm template --repo=https://SonarSource.github.io/helm-chart-sonarqube -s templates/secret.yaml -s templates/sonarqube-sts.yaml -s templates/jdbc-config.yaml --set jdbcOverwrite.enable=true --set jdbcOverwrite.jdbcUsername=sonar --set jdbcOverwrite.jdbcPassword=changeme --set postgresql.enabled=false sonarqube`

in your example there is a typo in `jdbcOverwrite.enabled`. it should be `jdbcOverwrite.enable`.

i can maybe also ensure you that our own production instance is not using a dummy or default password, it is running with a RDS database and gets deployed from this chart. i highly doubt that there is a logic flaw in that part in the chart.

tobias-trabelsi-sonarsource:
i will go ahead and close this PR now. If there is anything left, feel free to create a thread in our community forum or reopen this here

jef:
> in your example there is a typo in jdbcOverwrite.enabled. it should be jdbcOverwrite.enable.

Sorry, yes, that was a typo.

Okay, perhaps something strange was going on with our deployment. I'm going to do some further testing.

At the very least, I think the README verbiage should be changed. As:

> Enable JDBC overwrites for external Databases (disables `postgresql.enabled`)

is not correct. Would you agree? Since we've had to set `postgresql.enabled=false`.

jef:
@tobias-trabelsi-sonarsource, what do you think?

PR #113, opened 2021-11-05
Fix plugin download issue for StatefulSet chart

haljoh:
Downloading of plugins fail since the plugin init container is not allowed to write the downloaded plugin into the download directory. If setting values for plugins: runAsUser and runAsGroup to 0 the download works. But, when SonarQube container starts up it doesn't have the permissions to move the plugins in to the correct folder. 

This fix solve the issue so the default user-id 1000 can download the plugin and SonarQube container then have the permissions to move the downloaded files. This issue has been fixed in the deployment.yaml file but must have been missed for the stateful set.

### DIFF
--- a/charts/sonarqube/Chart.yaml
+++ b/charts/sonarqube/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: sonarqube
 description: SonarQube offers Code Quality and Code Security analysis for up to 27 languages. Find Bugs, Vulnerabilities, Security Hotspots and Code Smells throughout your workflow.
-version: 1.1.10
+version: 1.1.11
 appVersion: 9.1.0
 keywords:
   - coverage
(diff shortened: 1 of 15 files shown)


tobias-trabelsi-sonarsource (changes requested):
Hi @haljoh ,

thanks for your contribution. upon review i fully agree that the pod security context is missing from the stateful set. can you just add your change to the Chart.yaml description as well as the CHANGELOG.md and then we can merge your contribution

charts/sonarqube/Chart.yaml
```diff
@@ -21,10 +21,8 @@ annotations:
     - name: support
       url: https://community.sonarsource.com/
   artifacthub.io/changes: |
-    - kind: added
-      description: "added link to community support forum"
-    - kind: changed
-      description: "Use liveness endpoint instead of helth endpoint for liveness probe"
+    - kind: fixed
```
tobias-trabelsi-sonarsource, line 24: The previous changes have not been released yet, so please keep this list additive

tobias-trabelsi-sonarsource (changes requested):

charts/sonarqube/Chart.yaml
haljoh: Ahh, ok, shall I then also keep chart version as 1.1.10? I created a 1.1.11 version for my change. Should I also change that back to version 1.1.10?

tobias-trabelsi-sonarsource: nah the version increment to 1.1.11 (nice version btw :D) is correct

haljoh: ok, I have added back the changes to the Chart.yaml file

tobias-trabelsi-sonarsource:
The diff looks good to me now. sadly the git information seem off. could you try to squash your commits and rebase to master?

tobias-trabelsi-sonarsource (approved):

PR #116, opened 2021-11-15
DOCKER-17, DOCKER-29 and DOCKER-33

tobias-trabelsi-sonarsource:
DOCKER-29 Support ES auth in helm chart
DOCKER-17 Link to community forum as support channel
DOCKER-33 Switch liveness probe to the correct endpoint

### DIFF
--- a/charts/sonarqube-dce/README.md
+++ b/charts/sonarqube-dce/README.md
@@ -117,6 +117,11 @@ The following table lists the configurable parameters of the Sonarqube chart and
 | `searchNodes.image.pullPolicy` | search image pull policy | `IfNotPresent` |
 | `searchNodes.image.pullSecret` | search imagePullSecret to use for private repository | `nil` |
 | `searchNodes.env` | Environment variables to attach to the search pods | `nil` |
+| `searchNodes.searchAuthentication.enabled` | Securing the Search Cluster with basic authentication and TLS in between search nodes | `false` |
+| `searchNodes.searchAuthentication.keyStoreSecret` | Existing PKCS#11 Container as Keystore/Truststore to be used | `""` |
+| `searchNodes.searchAuthentication.keyStorePassword` | Password to Keystore/Truststore used in search nodes (optional) | `""` |
+| `searchNodes.searchAuthentication.keyStorePasswordSecret` | Existing secret for Password to Keystore/Truststore used in search nodes (optional) | `nil` |
+| `searchNodes.searchAuthentication.userPassword` | A User Password that will be used to authenticate against the Search Cluster | `""` |
 | `searchNodes.replicaCount` | Replica count of the Search Nodes | `3` |
 | `searchNodes.securityContext.fsGroup` | Group applied to mounted directories/files on search nodes | `1000` |
 | `searchNodes.containerSecurityContext.runAsUser` | User to run search container in sonarqube pod as | `1000` |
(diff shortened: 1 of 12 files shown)


charts/sonarqube-dce/README.md
```diff
@@ -117,6 +117,11 @@ The following table lists the configurable parameters of the Sonarqube chart and
 | `searchNodes.image.pullPolicy` | search image pull policy | `IfNotPresent` |
 | `searchNodes.image.pullSecret` | search imagePullSecret to use for private repository | `nil` |
 | `searchNodes.env` | Environment variables to attach to the search pods | `nil` |
+| `searchNodes.searchAuthentication.enabled` | Securing the Search Cluster with basic authentication and TLS in between search nodes | `false` |
+| `searchNodes.searchAuthentication.keyStoreSecret` | Existing Keystore/Truststore to be used | `""` |
+| `searchNodes.searchAuthentication.keyStorePassword` | Password to said Keystore/Truststore if set (optional) | `""` |
```
jacek-poreda-sonarsource, line 122: Why not just simply put : `Password to Keystore/Truststore used in search nodes (optional)` also it would be worth to mention what type of format we expect : `pkcs12`

This properties assumes that KeyStore and Truststore password is the same, though it could be different files with different password, is that intended?

charts/sonarqube-dce/README.md
```diff
@@ -117,6 +117,11 @@ The following table lists the configurable parameters of the Sonarqube chart and
...
+| `searchNodes.searchAuthentication.keyStorePasswordSecret` | existing secret for Password to said Keystore/Truststore (optional) | `nil` |
```
jacek-poreda-sonarsource, line 123: `Existing secret for Password to Keystore/Truststore (optional)`

charts/sonarqube-dce/templates/secret.yaml
```diff
@@ -14,32 +28,54 @@ data:
...
   labels:
     app: {{ template "sonarqube.name" . }}
     chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
     release: {{ .Release.Name }}
     heritage: {{ .Release.Service }}
 type: Opaque
 data:
-  SONAR_AUTH_JWTBASE64HS256SECRET: {{ .Values.ApplicationNodes.jwtSecret | b64enc | quote }}
+  SONAR_CLUSTER_ES_SSL_KEYSTOREPASSWORD: {{ .Values.searchNodes.searchAuthentication.keyStorePassword | b64enc | quote }}
+  SONAR_CLUSTER_ES_SSL_TRUSTSTOREPASSWORD: {{ .Values.searchNodes.searchAuthentication.keyStorePassword | b64enc | quote }}
```
jacek-poreda-sonarsource, line 63: As I said in previous comment, I think we should have different properties for keystore and truststore password

charts/sonarqube-dce/README.md
tobias-trabelsi-sonarsource: As long as we don't support `xpack.security.transport.ssl.verification_mode: full` there is no value in separating these two. it would increase complexity here without any gain. 
elastic is using them interchangeably in their documentation as well, so i would stick with using the same container for key and trust store for now. 

i will add a note about the format of the container tho (pkcs11 :wink: )

jacek-poreda-sonarsource (approved):

PR #121, opened 2021-11-26
docs: proposal to split up value table into sections of interest

k3rnelpan1c-dev:
### Description

I had this change for a time in a local branch, updated it and want to propose this change to make the table easier to read / find the values you are looking for.
It also gets rid of the many blank spaces in the Markdown Table which made it unnecessarily cumbersome to edit (IMHO).
An `.adoc` table would be even easier to edit from a markup perspective, yet that is even more subjective than this change already is, so I decided against proposing a markup language change.

Feedback welcome as always and I hope I didn't miss any options. 

### Changes

* split up values table into smaller section tables of interest

### Issues

* n/a?

### DIFF
--- a/charts/sonarqube-dce/README.md
+++ b/charts/sonarqube-dce/README.md
@@ -210,56 +210,115 @@ The following table lists the configurable parameters of the Sonarqube chart and
 
 | Parameter | Description | Default |
 | --------- | ----------- | ------- |
-| `OpenShift.enabled` | Define if this deployment is for OpenShift | `false` |
-| `OpenShift.createSCC` | If this deployment is for OpenShift, define if SCC should be created for sonarqube pod | `true` |
-| `elasticsearch.bootstrapChecks` | Enables/disables Elasticsearch bootstrap checks | `true` |
-| `nginx.enabled` | Also install Nginx Ingress Helm | `false` |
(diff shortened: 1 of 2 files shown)


tobias-trabelsi-sonarsource:
Hi @k3rnelpan1c-dev ,

this is definitely an improvement in terms of readability :+1:  we would need to do the same for the DCE chart tho. would you be willing to do the same for this chart as well or should i take care of this? :slightly_smiling_face:

k3rnelpan1c-dev:
Hi Tobias,

glad to hear you like this proposal :smiley:
Sure, I can do the same for the DEC chart too, likely this evening. I have only one question: Dose this documentation change alo require a version bump in the respective charts? (just so I can do it in one go)

tobias-trabelsi-sonarsource:
Thank you very much :pray: 
> Dose this documentation change also require a version bump in the respective charts?

nah i see this as a none functional change, so i would be fine with accepting these without a version bump :+1:

k3rnelpan1c-dev:
So, I hope I did not end up dropping or adding any values to the DEC Readme.
Let me know if I missed anything :)

tobias-trabelsi-sonarsource (approved):
LGTM. 
thank you very much for the improvement :D

PR #126, opened 2021-12-10
Add support for multiple image.pullSecrets

mrombout:
At the moment it is only possible to set one pull secret through `image.pullSecret`. But for my use-case I would like to be able to use multiple `imagePullSecrets` to be able to pull the images from different private registries.

Some changes along the lines of this PR would be required, would a feature like this to acceptable to merge?

---

In this PR I went for the easy way and replaced `image.pullSecret` with `image.pullSecrets` with a bit of effort it's also possible to implement this in a way that retains compatibility.

### DIFF
--- a/charts/sonarqube/Chart.yaml
+++ b/charts/sonarqube/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: sonarqube
 description: SonarQube offers Code Quality and Code Security analysis for up to 27 languages. Find Bugs, Vulnerabilities, Security Hotspots and Code Smells throughout your workflow.
-version: 1.2.5
+version: 2.0.0
 appVersion: 9.2.4
 keywords:
   - coverage
--- a/charts/sonarqube/Chart.yaml
+++ b/charts/sonarqube/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: sonarqube
 description: SonarQube offers Code Quality and Code Security analysis for up to 27 languages. Find Bugs, Vulnerabilities, Security Hotspots and Code Smells throughout your workflow.
-version: 2.0.0
+version: 1.3.0
 appVersion: 9.2.4
 keywords:
   - coverage
--- a/charts/sonarqube/Chart.yaml
+++ b/charts/sonarqube/Chart.yaml
@@ -23,6 +23,8 @@ annotations:
   artifacthub.io/changes: |
     - kind: added
       description: "added support multiple image pull secrets"
+    - kind: deprecated
+      description: "deprecated support for singular image pull secret"
     - kind: fixed
       description: "fixed missing image pull secret in admin hook job"
   artifacthub.io/containsSecurityUpdates: "false"
--- a/charts/sonarqube/Chart.yaml
+++ b/charts/sonarqube/Chart.yaml
@@ -21,8 +21,10 @@ annotations:
     - name: support
       url: https://community.sonarsource.com/
   artifacthub.io/changes: |
-    - kind: changed
-      description: "updated SonarQube to 9.2.4"
+    - kind: added
+      description: "added support multiple image pull secrets"
+    - kind: fixed
+      description: "fixed missing image pull secret in admin hook job"
   artifacthub.io/containsSecurityUpdates: "false"
   artifacthub.io/images: |
     - name: sonarqube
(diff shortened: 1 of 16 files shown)


tobias-trabelsi-sonarsource:
Hi @mrombout ,

thanks for the contribution. can you clarify why you need to have multiple pull secrets? usually you would have a separated deployment if you have separated staging/production registries/cluster where you could just overwrite the pull secret. i currently can not really think of a reason to define multiple.

also this is not a blocker for the LTS, so changing the LTS chart is out of question for this, but we surely can improve for the current chart as well as for the DCE chart.

mrombout:
None as of yet, but I intend on using this chart in a helmfile, so it's mainly the flexibility that I'm after. It's not too uncommon to have support for this (e.g. [influxdb](https://github.com/influxdata/helm-charts/tree/master/charts/influxdb), [postgresql](https://github.com/bitnami/charts/tree/master/bitnami/postgresql) (and other bitnami charts), etc.)

Having support multiple pull secrets would allow using `extraInitContainers` and `extraContainers` in environments that have the images for these containers on one private registry, and the DockerHub images on another private registry. And keep the secrets separate.

---

I've removed the changes to LTS and added the changes to DCE.

Since this is an incompatible change, should I change the major version. Or add back the old configuration parameters to keep compatibility and update the minor version?

tobias-trabelsi-sonarsource:
where would something like this be used? i am not convinced to merge a breaking change only based on a "this might be useful some day"  gut feeling tbh. a private registry should be a trust anchor and having multiple registries that are accessible from the same cluster used for the same deployment does not improve security while just adding complexity.  

If you can give me a clear example about the concerns i am open to reconsider.

larsla:
This would be useful for us since we want both authenticated access to Docker Hub (to work around the request quota) and being able to use an init container with an image from a private registry.

tobias-trabelsi-sonarsource:
@larsla what you can already do. `extraInitContainers` needs to be passed as a [container spec](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.21/#container-v1-core), so including a custom or additional pull secret

tobias-trabelsi-sonarsource:
Okay so not using a private registry as a cache for the docker hub was not on my radar as a valid use case, but it makes sense to reduce storage requirements for the private registry. authenticating against the docker hub to get higher rate limits also is a valid use case.

@mrombout the DCE chart is still in beta, so `0.3.0` seems appropriate. for the SonarQube chart as this is a potential breaking change, a major version bump is needed :+1:

tobias-trabelsi-sonarsource (changes requested):
Please mention your changes in the `CHANGELOG.md` as well as in the `Chart.yaml` so they are displayed on artifacthub.

also FYI: as this is a breaking change i pinged the responsible PM to make a final call on the versions. might take a little before we can merge

tobias-trabelsi-sonarsource (approved):
Functional review is okay.
Waiting for a decision from PM

tobias-trabelsi-sonarsource:
Hi @mrombout 
happy new year. I had a discussion in between the years with my PM and he states that this improvement can not be applied with a breaking change. If you can provide a backwards compatible solution you can deprecate `image.pullSecret` in favor of `image.pullSecrets` and mention this. what do you think?

tobias-trabelsi-sonarsource (changes requested):
Hi @mrombout 
happy new year. I had a discussion in between the years with my PM and he states that this improvement can not be applied with a breaking change. If you can provide a backwards compatible solution you can deprecate `image.pullSecret` in favor of `image.pullSecrets` and mention this. what do you think?

tobias-trabelsi-sonarsource:
uhm is this really ready for review @mrombout ? this still seems like a breaking change to me. if a user did previously define `image.pullSecret`, this would no longer work after this change.

charts/sonarqube/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: sonarqube
 description: SonarQube offers Code Quality and Code Security analysis for up to 27 languages. Find Bugs, Vulnerabilities, Security Hotspots and Code Smells throughout your workflow.
-version: 1.2.5
+version: 2.0.0
```
tobias-trabelsi-sonarsource, line 4: as this is no longer a breaking change a minor version bump should be enough

tobias-trabelsi-sonarsource (changes requested):

charts/sonarqube/Chart.yaml
tobias-trabelsi-sonarsource: same for the `CHANGELOG.md` else looks good :+1:

tobias-trabelsi-sonarsource (approved):
LGTM :+1: thanks for taking care of this :smiley:

PR #128, opened 2021-12-10
Use port's name

gbonnefille:
Use the port's name as its value is configurable.

### DIFF
--- a/charts/sonarqube-dce/CHANGELOG.md
+++ b/charts/sonarqube-dce/CHANGELOG.md
@@ -1,6 +1,9 @@
 # SonarQube Chart Changelog
 All changes to this chart will be documented in this file.
 
+## [0.2.2]
+* replaced hardcoded port's values
+
 ## [0.2.1]
 * updated SonarQube to 9.2.1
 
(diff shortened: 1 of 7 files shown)


tobias-trabelsi-sonarsource:
Hi @gbonnefille ,

thanks for your contribution and this is a good spotting of a potential error if a user defined sonarqube to be run on a different port. i would happily accept your PR if:
* you bump the chart version 
* do the fix for the sts and the dce as well 
* add your changes to the changelog

if using a named port is not possible (like in the bash script for the readinessProbe), you can use `{{.Values.service.internalPort}}`

tobias-trabelsi-sonarsource (changes requested):
Hi @gbonnefille ,

thanks for your contribution and this is a good spotting of a potential error if a user defined sonarqube to be run on a different port. i would happily accept your PR if:

* you bump the chart version
* do the fix for the sts and the dce as well
* add your changes to the changelog

if using a named port is not possible (like in the bash script for the readinessProbe), you can use `{{.Values.service.internalPort}}`

gbonnefille:
I did all required changes.

tobias-trabelsi-sonarsource (approved):
LGTM. Thanks :+1:

PR #138, opened 2021-12-22
Fix sonarqube-lts SCC

nfalco79:
Helm generates a wrong user entry in SCC that prefixes the specified service account name with Release.Name and does not matches the existing one
Example for serviceaccount default:
```
users:
- system:serviceaccount:namespace-x:sonarqube-lts-default
```
where the correct service account name is only default and not sonarqube-lts-default.
Expected:
```
users:
- system:serviceaccount:namespace-x:default
```

In sonarqube chart (non tls) seems already fixed in this defect fix (https://github.com/SonarSource/helm-chart-sonarqube/commit/fdd3e6d061387b01f4a36b0730f4d2578ea1afab)

### DIFF
--- a/charts/sonarqube-lts/templates/sonarqube-scc.yaml
+++ b/charts/sonarqube-lts/templates/sonarqube-scc.yaml
@@ -52,7 +52,7 @@ volumes:
 priority: 11
 users:
 {{- if .Values.serviceAccount.name }}
-- system:serviceaccount:{{ .Release.Namespace }}:{{ .Release.Name }}-{{ .Values.serviceAccount.name }}
+- system:serviceaccount:{{ .Release.Namespace }}:{{ .Values.serviceAccount.name }}
 {{- else  }}
 - system:serviceaccount:{{ .Release.Namespace }}:{{ .Release.Name }}-sonarqube
 {{- end }}


nfalco79:
@jacek-poreda-sonarsource could your review this PR. Changes are not different than those applied for defect DOCKER-19. I can confirm that it works on openshift 4.9

tobias-trabelsi-sonarsource:
Hi @nfalco79 ,

changes to the LTS chart are a bit tricky. You already have seen the work on DOCKER-19, so i guess you are aware of [the workaround](https://github.com/SonarSource/helm-chart-sonarqube/issues/104) to this issue? 

apart from that your PR needs a bit more work for the QA to pass (bump chart version and write changelog) and even then i can not guarantee that this will be merged, as there is a existing workaround

tobias-trabelsi-sonarsource:
closing due to inactivity

nfalco79:
> closing due to inactivity

what do you expect after this?

> and even then i can not guarantee that this will be merged, as there is a existing workaround

the workaround is create a service account that we already have so it's not a working workaround, I mean if I have to perform a dry-run and than apply changes manually heml is a bit useless.

For me does not make sense have a fix in all charts expect LTS that should be the most used for production...

tobias-trabelsi-sonarsource:
>what do you expect after this?

a response to my question and effort to make the PR pass the QA.

>the workaround is create a service account that we already have so it's not a working workaround

No. The workaround is to let the helm chart create the service account without an explicit name with values like this: 

```yaml
serviceAccount:
  create: true
  annotations: {}
```

nfalco79:
I need to use existing setup service account in existing project (namespace), if the workaround does not really generate a ServiceAccount yaml than it's ok. I will try with a --dry-run what it generates

tobias-trabelsi-sonarsource:
ah i misunderstood you then. 

if you already have a existing service account and want to use it, you can define the name as well and it will be picked up as well (see https://github.com/SonarSource/helm-chart-sonarqube/blob/master/charts/sonarqube-lts/templates/sonarqube-sts.yaml#L406 ). This will also create a new "wrong" Service account but you are free to delete this one. should not hinder your deployment and is rather a annoyance than a blocker

nfalco79:
the workaround generates:
```
---
# Source: sonarqube-lts/templates/sonarqube-scc.yaml
# This SCC allows any user ID but restricts capabilties and host access
apiVersion: security.openshift.io/v1
kind: SecurityContextConstraints
metadata:
  annotations:
    kubernetes.io/description: "allows pod to run as root, privileged and run sysctl"
    "helm.sh/hook": pre-install
  name: sonarqube-lts-privileged-scc
...
# If you want a priority on your SCC -- set for a value more than 0
priority: 11
users:
- system:serviceaccount:rad-jenkins:sonarqube-lts-sonarqube
MANIFEST:
---
# Source: sonarqube-lts/templates/serviceaccount.yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: sonarqube-lts-sonarqube-lts
---
```
As you can see heml will perform a service account creation and the scc refers to the wrong service account sonarqube-lts-sonarqube

tobias-trabelsi-sonarsource:
yeah you will need to alter the existing service account name. also please note that openshift is not official supported yet

nfalco79:
ok it's not official but I can confirm that with the fix and generated scc works.
Or better it requires also the following command from a openshift administrator account:
`oc adm policy add-scc-to-user sonarqube-lts-privileged-scc -z <service_account_name> -n <namespace>`

nfalco79:
I perform changed on changelog and Charts.yaml
I never used helm before so I do not know if something is missing. I take inspiration from other PR I saw here

PR #140, opened 2022-01-06
Add caCerts.enabled flag to Helm chart and README

jgallucci32:
The value `caCerts.enabled` is not specified in the helm chart. The Helm chart only looks for the existence of any value of `caCerts` being defined, such as `caCerts.secrets`. ~~This PR removes the unused field from the README example.~~

UPDATE: PR adds the flag properly to the Helm chart


tobias-trabelsi-sonarsource (changes requested):
I understand that this is confusing. may i propose instead of just deleting these lines, you mention this in the other values reference in the `README.md` and delete `caCerts.enabled`?

jgallucci32:
@tobias-trabelsi-sonarsource I do not mind going the other direction, however, if we go that way we should add the `caCerts.enabled` flag to the Helm chart rather than just use any value. This would make it more consistent with other parts of the chart.

tobias-trabelsi-sonarsource:
Which makes a lot of sense, but in that case you will need to bump the chart version and mention this change in the `Chart.yaml` as well as in the `CHANGELOG.md` file

tobias-trabelsi-sonarsource:
Hi @jgallucci32. I will close this PR due to inactivity while also implementing the suggested logic with #151

jgallucci32:
Sounds good, glad I was able to help in some way.

PR #146, opened 2022-01-12
Fix sonarqube-lts SCC

nfalco79:
Helm generates a wrong user entry in SCC that prefixes the specified service account name with Release.Name and does not matches the existing one
Example for serviceaccount default:
```
users:
- system:serviceaccount:namespace-x:sonarqube-lts-default
```
where the correct service account name is only default and not sonarqube-lts-default.
Expected:
```
users:
- system:serviceaccount:namespace-x:default
```

In sonarqube chart (non tls) seems already fixed in this defect fix (https://github.com/SonarSource/helm-chart-sonarqube/commit/fdd3e6d061387b01f4a36b0730f4d2578ea1afab)

### DIFF
--- a/charts/sonarqube-lts/CHANGELOG.md
+++ b/charts/sonarqube-lts/CHANGELOG.md
@@ -1,9 +1,12 @@
 # SonarQube Chart Changelog
 All changes to this chart will be documented in this file.
 
+## [1.0.25]
+* fixed wrong scc user reference if name was explicitly set
+
 ## [1.0.24]
 * fixed missing `env` key for the install-plugins container in both the Deployment and StatefulSet
-  
+
 ## [1.0.23]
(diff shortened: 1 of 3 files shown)


nfalco79:
@tobias-trabelsi-sonarsource this PR in place of #138

tobias-trabelsi-sonarsource:
i will ask my PM if this defect justifies a LTS change. in the mean time please state your change in `artifacthub.io/changes` as well (`Chart.yaml`)

tobias-trabelsi-sonarsource:
After discussing it with my PM he came back with the same message that i already gave you. 

the workaround covers the serviceaccount that is created by the helm chart. if you want to use your own, this is not yet supported. you can benefit from DOCKER-19 if you are using the latest version of the chart/sonarqube.

PR #158, opened 2022-01-27
Add aml Github hook

sousadax12:


charts/sonarqube/templates/_helpers.tpl
```diff
@@ -154,4 +154,17 @@ Create the name of the service account to use
...
 {{- end -}}
-{{- end -}}
\ No newline at end of file
+{{- end -}}
+
+{{/*
+Create curl request to cconfigure aml github
+*/}}
+{{- define "sonarqube.aml.github"  -}}
+curl -L -k -u admin:{{ .Values.account.adminPassword | default "admin" }} -X POST "{{ template "sonarqube.fullname" . }}:{{ default 9000 .Values.service.internalPort }}/api/alm_settings/create_github" \
```
tobias-trabelsi-sonarsource, line 163: This should honor a custom web context

charts/sonarqube/templates/apply-aml-github-job.yaml
```diff
@@ -0,0 +1,45 @@
+{{- if .Values.aml }}
+{{- if .Values.aml.github }}
+{{- if and .Values.aml.github.appId .Values.aml.github.url .Values.aml.github.clientId .Values.aml.github.clientSecret .Values.aml.github.key .Values.aml.github.privateKey }}
```
tobias-trabelsi-sonarsource, line 3: i understand why you want to catch everything but following the principals from the rest of this helm chart a simple `enable` flag should be enough. 
A failing container is more easy to debug than a job that is never created in the first place if a user in the future does not provide all required values

charts/sonarqube/templates/apply-aml-github-job.yaml
```diff
@@ -0,0 +1,45 @@
...
+kind: Job
+metadata:
+  name: {{ template "sonarqube.fullname" . }}-apply-aml-github-hook
+  labels:
+    app: {{ template "sonarqube.name" . }}
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+    helm.sh/chart: "{{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}"
+  annotations:
+    "helm.sh/hook": post-install,post-upgrade
```
tobias-trabelsi-sonarsource, line 14: i think this would do harm if ran after upgrade. The DevOpsPlatform settings (previously known as ALM) persist in the database and do not need to be re added after a upgrade

charts/sonarqube/templates/apply-aml-github-job.yaml
```diff
@@ -0,0 +1,45 @@
...
+        release: {{ .Release.Name }}
+      {{- range $key, $value := .Values.service.labels }}
+        {{ $key }}: {{ $value | quote }}
+      {{- end }}
+    spec:
+      restartPolicy: OnFailure
+      containers: 
+      - name: {{ template "sonarqube.fullname" . }}-apply-aml-github
+        image: {{ default "curlimages/curl:latest" .Values.curlContainerImage }}
+        {{- if $securityContext := (.Values.initContainers.securityContext) }}
```
tobias-trabelsi-sonarsource, line 33: reusing the `initContainers` scope is a bit misleading as this is not a init container

charts/sonarqube/values.yaml
```diff
@@ -500,3 +500,13 @@ extraConfig:
 # sonarWebContext: /
 
 terminationGracePeriodSeconds: 60
+
+# aml:
```
tobias-trabelsi-sonarsource, line 504: the term used to be ALM as in application live-cycle management, but is now going in the application as "DevOps Platforms"

tobias-trabelsi-sonarsource (changes requested):
Hi @sousadax12 ,

thanks for creating this PR. i have added some remarks to it that definitely needs to be addressed before taking this into account any further. 
While i understand your approach here, i am not sure if this is the right way to do it. If following a config as code principle for sonarqube, it should be done in the application and not worked around in this helm chart. I will discuss this internally and come back to you in a while

charts/sonarqube/templates/apply-aml-github-job.yaml
sousadax12: I decided to add on post-install and post-upgrade because this way these values can be changed using a pipeline or something like that without a fresh install.

sousadax12:
> 

Thank you for the review 👍 
My idea here was to give the possibility to configured everything using code. I like to have that possibility, like using terraform to configure the Github Application and using the output values to deploy the chart. This will allow us to have configuration as code, avoiding mistakes.

tobias-trabelsi-sonarsource:
Hi @sousadax12 ,

after discussing this issue with my PM he confirmed that this should not be worked around in the helm chart. I welcome you to raise your point on [Prductboard](https://portal.productboard.com/sonarsource/3-sonarqube/tabs/12-planned-for-9-4) an/or on our [community forum](https://community.sonarsource.com) where somebody else already created [a thread about it](https://community.sonarsource.com/t/feature-request-allow-a-way-to-bootstrap-initial-configuration-sonarqube/48860). 
Thank you again for providing a PR and sorry for not accepting this

PR #159, opened 2022-01-28
DOCKER-49 only one (Properties or SecretProperties) is needed

### LINKED ISSUE #135: sonarSecretProperties only used if sonarProperties also specified
## Description
I noticed something when installing SonarQube with the helm chart. If you only specify `sonarSecretProperties` when deploying and don't specify any `sonarProperties`; then none of the `sonarSecretProperties` get used. I didn't try it in the other direction (only specifying `sonarProperties`), but I would imagine that behaves the same way. I say this because I noticed that both the [Deployment](https://github.com/SonarSource/helm-chart-sonarqube/blob/master/charts/sonarqube/templates/deployment.yaml#L141) and [StatefulSet](https://github.com/SonarSource/helm-chart-sonarqube/blob/master/charts/sonarqube/templates/sonarqube-sts.yaml#L111) seem to run those properties through an `and` conditional.

## Workaround
I was able to work around my issue but just specifying at least one property in `sonarProperties`; that way both had some values specified and thus all of the properties were imported.

## Expected Behavior
I would expect that only one of those parameters needs specifying, not both.

tobias-trabelsi-sonarsource:
closes #135

### DIFF
--- a/charts/sonarqube-dce/CHANGELOG.md
+++ b/charts/sonarqube-dce/CHANGELOG.md
@@ -1,6 +1,9 @@
 # SonarQube Chart Changelog
 All changes to this chart will be documented in this file.
 
+## [0.6.4]
+* properties are now correctly set
+
 ## [0.6.3]
 * `livenessProbe.failureThreshold` was never rendered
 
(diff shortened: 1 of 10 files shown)


charts/sonarqube-dce/CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 # SonarQube Chart Changelog
 All changes to this chart will be documented in this file.
 
+## [0.6.1]
```
pierre-guillot-sonarsource, line 4:
```suggestion
## [0.6.4]
```
Should be 0.6.4, right ?

tobias-trabelsi-sonarsource: yes indeed :see_no_evil:

pierre-guillot-sonarsource (approved):

PR #163, opened 2022-01-31
REL-2147 DOCKER-42 Update SonarQube to 9.3.0; officially support DCE Chart

wouter-admiraal-sonarsource:


charts/sonarqube/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v1
 name: sonarqube
 description: SonarQube offers Code Quality and Code Security analysis for up to 27 languages. Find Bugs, Vulnerabilities, Security Hotspots and Code Smells throughout your workflow.
-version: 1.6.5
-appVersion: 9.2.4
+version: 1.6.6
```
tobias-trabelsi-sonarsource, line 4: i don't think an application version bump is a patch release :sweat_smile: 
as there are a few changes in this release that are user facing, i think this even justifies a major version bump. wdyt?

tobias-trabelsi-sonarsource (changes requested):
You need to change the `values.yaml` as well to reference the new images

charts/sonarqube/Chart.yaml
wouter-admiraal-sonarsource: Works for me 👍

tobias-trabelsi-sonarsource (approved):
LGTM :tada:

PR #171, opened 2022-02-09
Updated all instances of the caCerts enabled check

kevin-utilus:
Not all instances of the check if caCerts were enabled were updated so it would still try to add the volumes even though it is disabled. This causes the deployment to fail.


tobias-trabelsi-sonarsource (approved):
LGTM.
Thank you very much for your contribution @kevin-utilus

PR #180, opened 2022-02-21
Fix spelling error in sonarqube-lts App sts: "initFS.enabeld" -> "initFs.enabled"

danragnar:
- [X] Bump the Version number of the respected chart

Fixes spelling error in the App Statefulset for enabling or disabling the initFs initContainer.

### DIFF
--- a/charts/sonarqube-lts/templates/sonarqube-sts.yaml
+++ b/charts/sonarqube-lts/templates/sonarqube-sts.yaml
@@ -161,7 +161,7 @@ spec:
             {{- . | toYaml | trim | nindent 12 }}
           {{- end }}
       {{- end }}
-      {{- if and .Values.persistence.enabled .Values.initFs.enabled }}
+      {{- if and .Values.persistence.enabled ( or .Values.initFs.enabeld .Values.initFs.enabled ) }}
         - name: init-fs
           image: {{ default "busybox:1.32" .Values.initFs.image }}
           imagePullPolicy: {{ .Values.image.pullPolicy  }}
--- a/charts/sonarqube-lts/templates/sonarqube-sts.yaml
+++ b/charts/sonarqube-lts/templates/sonarqube-sts.yaml
@@ -161,7 +161,7 @@ spec:
             {{- . | toYaml | trim | nindent 12 }}
           {{- end }}
       {{- end }}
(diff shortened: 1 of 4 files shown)


charts/sonarqube-lts/templates/sonarqube-sts.yaml
```diff
@@ -161,7 +161,7 @@ spec:
             {{- . | toYaml | trim | nindent 12 }}
           {{- end }}
       {{- end }}
-      {{- if and .Values.persistence.enabled .Values.initFs.enabeld }}
+      {{- if and .Values.persistence.enabled .Values.initFs.enabled }}
```
wouter-admiraal-sonarsource, line 164: In order to keep the previous "faulty" value for backwards compatibility, how about something like (to be adapted):

```suggestion
      {{- if and .Values.persistence.enabled ( or .Values.initFs.enabeld .Values.initFs.enabled ) }}
```

wouter-admiraal-sonarsource (changes requested):
Hi @danragnar , thank you very much for your contribution.

Unfortunately, this is a breaking change. We would have to make it backward compatible. I made a few suggestion.

danragnar:
> Hi @danragnar , thank you very much for your contribution.
> 
> Unfortunately, this is a breaking change. We would have to make it backward compatible. I made a few suggestion.

Yeah, seems like a good idea. I have added your proposed changes.

wouter-admiraal-sonarsource:
🤦  I just realized this is targetting our LTS chart. This is contrary to our policy: we only provide updates if there's a security vulnerability or a major degradation in functionality. A typo in a configuration file, although kind of dumb, is not a valid reason to update our chart.

I'm really sorry I didn't notice this earlier. I'm going to have to close this PR. Thank you again for taking the time to open a PR and to attempt to fix this. It's really appreciated.

danragnar:
I understand. But what you can do is fix the documentation so that when someone wants to disable the initFs initcontainer, they're not set out on a 2 hour troubleshooting extravaganza as I was...

wouter-admiraal-sonarsource:
That's a fair point 👍 . Updated it here: https://github.com/SonarSource/helm-chart-sonarqube/blob/master/charts/sonarqube-lts/README.md

PR #182, opened 2022-02-28
Secure admin password in k8s secret

kpink224:

----
While the change admin password hook is very handy, we noticed it exposes the admin password in the deploy logs. For security reasons, we need this password to be stored as a k8s secret.

### DIFF
--- a/charts/sonarqube/templates/secret.yaml
+++ b/charts/sonarqube/templates/secret.yaml
@@ -26,3 +26,21 @@ metadata:
 type: Opaque
 data:
   SONAR_WEB_SYSTEMPASSCODE: {{ .Values.monitoringPasscode | b64enc | quote }}
+---
+{{- if .Values.account }}
+{{- if .Values.account.adminPassword }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ template "sonarqube.fullname" . }}-admin-password
+  labels:
+    app: {{ template "sonarqube.name" . }}
+    chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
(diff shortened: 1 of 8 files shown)


charts/sonarqube/templates/secret.yaml
```diff
@@ -26,3 +26,21 @@ metadata:
...
+metadata:
+  name: {{ template "sonarqube.fullname" . }}-admin-password
+  labels:
+    app: {{ template "sonarqube.name" . }}
+    chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+type: Opaque
+stringData:
+  password: {{ .Values.account.adminPassword | urlquery }}
```
kpink224, line 43: The only real difference is the new admin password doesn't have a default since `change-admin-password-hook.yml` is only applied if the new admin password is defined.

kpink224:
Tested this implementation is a generic password (i.e. [0-9A-Z]) as well as url encoded values: `#$hello/`

tobias-trabelsi-sonarsource (changes requested):
Hi @kpink224 ,

thanks you very much for your contribution. indeed a bump is necessary as we use and test everything internally before releasing to the public. 
Could you also port your changes to the DCE chart? i think this is a very good improvement to have :smiley:

kpink224:
@tobias-trabelsi-sonarsource Made the changes as requested. Lmk if you need anything else!

kpink224:
@tobias-trabelsi-sonarsource Edited that left over syntax issue

tobias-trabelsi-sonarsource (approved):
LGTM

PR #183, opened 2022-03-03
Specify service account name in change admin password hook

kpink224:

----
We noticed in our gatekeeper constraints that the admin password hook was using the default service account even though we specified a service account to be used. This change just uses that specified service account.


tobias-trabelsi-sonarsource:
good catch!

tobias-trabelsi-sonarsource (approved):

PR #187, opened 2022-03-11
Specify location of .netrc file if netrcCreds are provided for plugin installation.

Tobytee17:
**Description**
There is currently an issue where if a user needs to provide credentials to a server in order to retrieve a plugin for installation, the .netrc file that is created by the chart (using the secret specified at `.Values.plugins.netrcCreds`) is not used. This causes issues in scenarios where users want to install internal plugins that have been published to private artifact registries that don't allow anonymous access.

Normally we would just need to specify the `'-n'` flag to curl to tell it to use the .netrc file in the users home directory, but as we have mounted the .netrc secret at the root directory (see [here](https://github.com/SonarSource/helm-chart-sonarqube/blob/master/charts/sonarqube/templates/deployment.yaml#L127)), we have to specify the `'--netrc-file'` flag to tell it to retrieve the file from the /root directory.

**Changes**

- Added conditional rendering of the `--netrc-file` flag (driven by the value provided in `.Values.plugins.netrcCreds`), and have defaulted the value of it to /root/.netrc. I've not made the location for this configurable, as the mount location for the secret mentioned above is also not configurable, so its not really expected to change.

### DIFF
--- a/charts/sonarqube-dce/CHANGELOG.md
+++ b/charts/sonarqube-dce/CHANGELOG.md
@@ -1,6 +1,9 @@
 # SonarQube Chart Changelog
 All changes to this chart will be documented in this file.
 
+## [1.0.7]
+* Specify location of .netrc file when downloading plugins that require auth
+
 ## [1.0.6]
 * Fixed properties scope for app deployment and search sts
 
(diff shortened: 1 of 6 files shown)


tobias-trabelsi-sonarsource (changes requested):
Hi @Tobytee17 ,
thanks for your contribution. can you add them to the `Chart.yaml` file (`artifacthub.io/changes`) as well and port them to the DCE Chart? I think it is affected by the same bug

Tobytee17:
Ah sorry, totally missed the annotation when looking at the Chart.yml. I've made the required changes, and have ported them to the DCE chart as well.

tobias-trabelsi-sonarsource (approved):
LGTM, Thanks again :smiley:

PR #196, opened 2022-03-29
DOCKER-64 updated Chart.yaml api version

### LINKED ISSUE #194: Update Chart.yaml apiVersion to v2
**Describe the bug**
Since helm3 is the only supported version the chart should use apiVersion: v2 to reflect that. This is documented here https://helm.sh/docs/topics/charts/#the-apiversion-field

**Versions used:**
- Chart Version: 2.0.7
- SonarQube Version: 9.3.0
- Helm Version: helm3
- Kubernetes Version: v1.21

**Expected behavior**
apiVersion should be v2

Lerentis:
closes #194

### DIFF
--- a/charts/sonarqube-dce/CHANGELOG.md
+++ b/charts/sonarqube-dce/CHANGELOG.md
@@ -1,6 +1,10 @@
 # SonarQube Chart Changelog
 All changes to this chart will be documented in this file.
 
+
+## [1.0.8]
+* updated helm api version to make helm3 explicit
+
 ## [1.0.7]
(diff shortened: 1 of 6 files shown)


leo-geoffroy-sonarsource:
Thanks for your contribution @Lerentis :)
Do we foresee any possible issue with  LTS?

Lerentis:
I don't see any. Helm 3 was the only supported version of the chart from the very beginning

Lerentis:
Closing this pr as it does not seem to be fitting into the new pr policy that was introduced with #199

PR #202, opened 2022-04-01
Add ingressClassName for LTS

shizacat:
Add ingressClassName for LTS


leo-geoffroy-sonarsource:
Hi @shizacat 
Sadly this change cannot qualify for the LTS version as it is not a blocker.
Please use the latest version if you need this configuration

cdaniluk:
@leo-geoffroy-sonarsource `ingressClassName` is required for K8s 1.22 and higher. 1.22 is the oldest supported release (and is about to go out of support in a few months). I don't think the "blocker only" policy works with the concept of LTS if it prevents you from actually running a supported configuration.

PR #231, opened 2022-06-10
Fix Openshift route name

jonesbusy:
Hi,

Name of the Openshift/OKD should use fullname instead of name like other resources.

Thanks!


jonesbusy:
Thanks for resolving the merge conflict due to 4.0.0 release!

leo-geoffroy-sonarsource (approved):

leo-geoffroy-sonarsource:
Hi @jonesbusy 

Thanks for the contribution and good catch! I think it makes sense to use fullName here indeed
I have merged the PR

PR #240, opened 2022-06-29
BUILD-1675 Remove helm-index GitHub action from gh-action_release

drautureau-sonarsource:
- [x] explain your motives to contribute this change: what problem you are trying to fix, what improvement you are trying to make: **See https://jira.sonarsource.com/browse/BUILD-1675**
- [x] Document your Changes in the `CHANGELOG.md` file of the respected chart as well as the `Chart.yaml`:  **Not required**
- [x] Bump the Version number of the respected chart: **Not required**

**This was successfully tested on this fork: https://github.com/drautureau-sonarsource/helm-chart-sonarqube**

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -29,12 +29,14 @@ jobs:
       - name: Create local repository directory
         id: local_repo
         run: echo ::set-output name=dir::"$(mktemp -d repo.XXXXXXXX)"
-      - name: Download Artifacts
-        uses: SonarSource/gh-action_release/download-build@v4
-        with:
-          build-number: ${{ github.event.inputs.buildNumber }}
-          remote-repo: 'sonarsource-helm'
-          local-repo-dir: ${{ steps.local_repo.outputs.dir }}
+      - name: Download artifacts
+        shell: bash
+        working-directory: ${{ steps.local_repo.outputs.dir }}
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -29,12 +29,14 @@ jobs:
       - name: Create local repository directory
         id: local_repo
         run: echo ::set-output name=dir::"$(mktemp -d repo.XXXXXXXX)"
-      - name: Download Artifacts
-        uses: SonarSource/gh-action_release/download-build@v4
-        with:
-          build-number: ${{ github.event.inputs.buildNumber }}
-          remote-repo: 'sonarsource-helm'
-          local-repo-dir: ${{ steps.local_repo.outputs.dir }}
```
malena-ebert-sonarsource, line 37: Why do you remove this part and duplicated the code?

It is a generic action to download full builds from repox.

drautureau-sonarsource: This was the root cause of the issue https://github.com/SonarSource/helm-chart-sonarqube/runs/6876350938?check_suite_focus=true

Sometimes, duplicating some code from elsewhere removes the complexity the dependency is introducing
And in this case, it is at the end 2 added lines of code:


I also believe that the download action is not a core feature of the release process (and the code relies on external tooling from Jfrog).

PR #243, opened 2022-08-15
BUILD-1808 update repox deployer credentials

tomverin:


wouter-admiraal-sonarsource:
This still doesn't seem to work. Maybe it's another value?

tomverin:
> This still doesn't seem to work. Maybe it's another value?

I messed up the fix, went to fast and missed another occurence of the credentials, fixing the fix

wouter-admiraal-sonarsource (approved):

PR #244, opened 2022-08-17
Made StatefulSet replicas configurable

mloz-sbv:
This will allow to scale down a stateful set to 0 using gitops.

### DIFF
--- a/charts/sonarqube-lts/CHANGELOG.md
+++ b/charts/sonarqube-lts/CHANGELOG.md
@@ -1,6 +1,9 @@
 # SonarQube Chart Changelog
 All changes to this chart will be documented in this file.
 
+## [1.0.30]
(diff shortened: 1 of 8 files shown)


mloz-sbv:
Thanks for the feedback @carminevassallo. I like the added validation! I'd just like to mention the lts chart still has a [fixed ](https://github.com/SonarSource/helm-chart-sonarqube/blob/master/charts/sonarqube-lts/templates/sonarqube-sts.yaml#L18)`replicaCount` value though.

Best regards,

Sven

carminevassallo:
Hi @mloz-sbv,

yes, we know and this is intentional. We are gonna update the lts chart in a few months right after releasing the new sonarqube lts version.

Regards,
Carmine

PR #246, opened 2022-08-18
DOCKER-77 Add documentation for ingress tls in helm charts

carminevassallo:
I add a reference about generating self-signed certificates (and private keys) and registering the corresponding secret in the cluster.


leo-geoffroy-sonarsource:
The new documentation looks good!
Since the chart version has changed, you can also update the README to add a line to the changelog 
You can also squash the two commits into a single one

leo-geoffroy-sonarsource (changes requested):
readme changes and squash commits

leo-geoffroy-sonarsource (approved):
LGTM

PR #247, opened 2022-08-19
DOCKER-63 Set the number of allowed replicas in sonarqube and sonarqube-lts charts

carminevassallo:
We only support 1 replica for deployment and statefulset types in sonarqube and sonarqube-lts charts. Therefore, we set 0 and 1 as allowed values for `replicas`, having 1 as the default value.

### DIFF
--- a/charts/sonarqube/CHANGELOG.md
+++ b/charts/sonarqube/CHANGELOG.md
@@ -1,6 +1,9 @@
 # SonarQube Chart Changelog
 All changes to this chart will be documented in this file.
 
+## [5.0.2]
+* Set the number of allowed replicas to 0 and 1
+
 ## [5.0.1]
 * Add documentation for ingress tls
 
--- a/charts/sonarqube/Chart.yaml
+++ b/charts/sonarqube/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: sonarqube
 description: SonarQube offers Code Quality and Code Security analysis for up to 27 languages. Find Bugs, Vulnerabilities, Security Hotspots and Code Smells throughout your workflow.
-version: 5.0.1
+version: 5.0.2
 appVersion: 9.6.0
 keywords:
   - coverage
--- a/charts/sonarqube/README.md
+++ b/charts/sonarqube/README.md
@@ -118,7 +118,7 @@ The following table lists the configurable parameters of the Sonarqube chart and
 | Parameter | Description | Default |
 | --------- | ----------- | ------- |
 | `deploymentType` | Deployment Type (supported values are `StatefulSet` or `Deployment`) | `StatefulSet` |
-| `replicaCount`   | Number of replicas deployed  | `1` |
+| `replicaCount`   | Number of replicas deployed (supported values are 0 and 1)  | `1` |
 | `deploymentStrategy` | Deployment strategy | `{}` |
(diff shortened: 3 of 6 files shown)


charts/sonarqube/templates/deployment.yaml
```diff
@@ -15,7 +15,11 @@ metadata:
     app.kubernetes.io/component: {{ template "sonarqube.fullname" . }}
     app.kubernetes.io/version: {{ tpl .Values.image.tag . | quote }}
 spec:
-  replicas: {{ .Values.replicaCount }}
+  {{- if or (eq .Values.replicaCount 1.0) (eq .Values.replicaCount 0.0)}}
```
leo-geoffroy-sonarsource, line 18: Is there a specific limitation in helm that prevents us to use integer instead of double here?

carminevassallo: the `replicaCount` value is recognized as float, therefore I was setting 1.0 in the comparison. I can typecast the value and have something like `(eq (int .Values.replicaCount) 1) ` in the condition. Would that be acceptable for you? :)

carminevassallo: We discussed this privately and decided to remove such a "behind-the-scene" reset of the `replicaCount` value, introducing input validation for `replicaCount` (credits to [this blog post](https://austindewey.com/2020/06/13/helm-tricks-input-validation-with-values-schema-json/#:~:text=The%20values.,annotate%20and%20validate%20JSON%20documents%E2%80%9D) by Austin Dewey). The solution will be detailed in the next commit.

carminevassallo:
We introduce the input validation for `replicaCount`. Now users specifying a value different from 0 and 1, will get the following error: 
```
Error: UPGRADE FAILED: values don't meet the specifications of the schema(s) in the following chart(s):
sonarqube:
- replicaCount: replicaCount must be one of the following: 0, 1
```
We also simulated the following upgrade scenario:
- Kathy is running version 5.0.1 (it does not include input validation) with `replicaCount` set to 2 (although this is not recommended, the 2 pods are up and running)
- Kathy tries to upgrade to version 5.0.2 (with input validation) using the same `values.yaml`. The upgrade fails with the above error message. This happens also if they try to set `replicaCount` via command line (`--set replicaCount=2`)

Which are your thoughts on that @leo-geoffroy-sonarsource ?

leo-geoffroy-sonarsource:
> We introduce the input validation for `replicaCount`. Now users specifying a value different from 0 and 1, will get the following error:
> 
> ```
> Error: UPGRADE FAILED: values don't meet the specifications of the schema(s) in the following chart(s):
> sonarqube:
> - replicaCount: replicaCount must be one of the following: 0, 1
> ```
> 
> We also simulated the following upgrade scenario:
> 
> * Kathy is running version 5.0.1 (it does not include input validation) with `replicaCount` set to 2 (although this is not recommended, the 2 pods are up and running)
> * Kathy tries to upgrade to version 5.0.2 (with input validation) using the same `values.yaml`. The upgrade fails with the above error message. This happens also if they try to set `replicaCount` via command line (`--set replicaCount=2`)
> 
> Which are your thoughts on that @leo-geoffroy-sonarsource ?

Looks good to me! I tested it locally and everything works as expected
I think we are good to go after a commit squash :)

leo-geoffroy-sonarsource (approved):

PR #249, opened 2022-08-23
DOCKER-78 fix wrong conditions for sonarProperties and sonarSecretProperties and documentation update

carminevassallo:
Based on my current understanding, we have 4 ways to specify these properties:

- Rely on the default `sonar.properties` already available inside the containers
- Mount your own `sonar.properties`
- Mount your own `secret.properties` (from a secret spec)
- A combination of 3 and 2 

In the DCE chart, I think that we cover correctly only case 1. We could probably refactor the code in this way to achieve the other cases.

### DIFF
--- a/charts/sonarqube-dce/README.md
+++ b/charts/sonarqube-dce/README.md
@@ -166,8 +166,8 @@ The following table lists the configurable parameters of the Sonarqube chart and
 | `ApplicationNodes.image.pullSecret` | (DEPRECATED) app imagePullSecret to use for private repository | `nil` |
 | `ApplicationNodes.image.pullSecrets` | app imagePullSecrets to use for private repository | `nil` |
 | `ApplicationNodes.env` | Environment variables to attach to the app pods | `nil` |
-| `ApplicationNodes.sonarProperties` | Custom `sonar.properties` file for App Nodes | `None` |
-| `ApplicationNodes.sonarSecretProperties` | Additional `sonar.properties` file for App Nodes to load from a secret | `None` |
+| `ApplicationNodes.sonarProperties` | Custom `sonar.properties` key-value pairs for App Nodes (e.g., "ApplicationNodes.sonarProperties.sonar.forceAuthentication=true") | `None` |
+| `ApplicationNodes.sonarSecretProperties` | Additional `sonar.properties` key-value pairs for App Nodes to load from a secret | `None` |
 | `ApplicationNodes.sonarSecretKey` | Name of existing secret used for settings encryption | `None` |
 | `ApplicationNodes.replicaCount` | Replica count of the app Nodes | `2` |
 | `ApplicationNodes.podDistributionBudget` | PodDisctributionBudget for the App Nodes | `minAvailable: "50%"` |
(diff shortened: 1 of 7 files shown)


leo-geoffroy-sonarsource (changes requested):
Check DOCKER-78 for changes

carminevassallo:
I've just tested all three cases when using the DCE chart. I've removed the unreachable condition and I've tested it anyway (despite not making too much sense xD)...

charts/sonarqube/README.md
```diff
@@ -287,8 +287,8 @@ The following table lists the configurable parameters of the Sonarqube chart and
 | `jvmOpts` | Values to add to SONARQUBE_WEB_JVM_OPTS | `""` |
 | `jvmCeOpts` | Values to add to SONAR_CE_JAVAOPTS | `""` |
 | `sonarqubeFolder` | Directory name of Sonarqube | `/opt/sonarqube` |
-| `sonarProperties` | Custom `sonar.properties` file | `None` |
-| `sonarSecretProperties` | Additional `sonar.properties` file to load from a secret | `None` |
+| `sonarProperties` | Custom `sonar.properties` values (e.g., "sonarProperties.sonar.forceAuthentication=true") | `None` |
```
leo-geoffroy-sonarsource, line 290: I would use "key-value pairs" here instead of values to align on the two properties

charts/sonarqube-dce/README.md
```diff
@@ -166,8 +166,8 @@ The following table lists the configurable parameters of the Sonarqube chart and
 | `ApplicationNodes.image.pullSecret` | (DEPRECATED) app imagePullSecret to use for private repository | `nil` |
 | `ApplicationNodes.image.pullSecrets` | app imagePullSecrets to use for private repository | `nil` |
 | `ApplicationNodes.env` | Environment variables to attach to the app pods | `nil` |
-| `ApplicationNodes.sonarProperties` | Custom `sonar.properties` file for App Nodes | `None` |
-| `ApplicationNodes.sonarSecretProperties` | Additional `sonar.properties` file for App Nodes to load from a secret | `None` |
+| `ApplicationNodes.sonarProperties` | Custom `sonar.properties` values for App Nodes (e.g., "ApplicationNodes.sonarProperties.sonar.forceAuthentication=true") | `None` |
```
leo-geoffroy-sonarsource, line 169: I would use key-value pairs instead of values to align on the two topics

leo-geoffroy-sonarsource (changes requested):
A little change and maybe we could rename the PR to reflect what has really been done (fix conditions and documentation)

Also I tested, and all the scenarios are working as expected :+1:

charts/sonarqube/README.md
carminevassallo: yes it makes sense

charts/sonarqube-dce/README.md
carminevassallo: yes it makes sense

leo-geoffroy-sonarsource (approved):

PR #250, opened 2022-08-24
DOCKER-64 bump apiVersion to v2 and refactor deprecated configs

carminevassallo:
We have to update our `apiVersion` to `v2`, as this is the recommended API used by Helm 3 (which is the version that we support).

We checked the official [migration guidelines](https://helm.sh/docs/topics/v2_v3_migration/) and make our chart complaint  with the following changes in `apiVersion=v2`:
- "Dynamically linked chart dependencies moved to Chart.yaml (requirements.yaml removed and requirements --> dependencies)"
- "Charts have a type metadata field to define the chart to be of an application or library chart. It is application by default which means it is renderable and installable"

### DIFF
--- a/charts/sonarqube-dce/CHANGELOG.md
+++ b/charts/sonarqube-dce/CHANGELOG.md
@@ -1,6 +1,9 @@
 # SonarQube Chart Changelog
 All changes to this chart will be documented in this file.
 
+## [4.0.3]
+* Bump apiVersion to v2
+
 ## [4.0.2]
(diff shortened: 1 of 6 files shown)


leo-geoffroy-sonarsource:
Tested and everything work as expected!
Nitpick : we are missing a dash in the commit message (DOCKER-64)

leo-geoffroy-sonarsource (changes requested):
small adjustment in the commit message

leo-geoffroy-sonarsource (approved):

PR #251, opened 2022-08-25
DOCKER-80 add the possibility to use an existing secret for change admin password hook

carminevassallo:
Currently, users cannot define their own secret in the cluster for setting the current and custom passwords at the startup. This PR aims at providing our standard and dce chart users with this possibility.

### DIFF
--- a/charts/sonarqube-dce/values.yaml
+++ b/charts/sonarqube-dce/values.yaml
@@ -549,8 +549,12 @@ extraConfig:
   configmaps: []
 
 # account:
+# The values can be set to define the current and the (new) custom admin passwords at the startup (the username will remain "admin")
 #   adminPassword: admin
 #   currentAdminPassword: admin
+# The above values can be also provided by a secret that contains "password" and "currentPassword" as keys. You can generate such a secret in your cluster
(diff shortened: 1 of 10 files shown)


charts/sonarqube-dce/values.yaml
```diff
@@ -549,8 +549,11 @@ extraConfig:
   configmaps: []
 
 # account:
+# The values can be set to define the current and the (new) custom admin passwords at the startup (the username will remain "admin")
 #   adminPassword: admin
 #   currentAdminPassword: admin
+# The above values can be also provided by a secret that contains "password" and "currentPassword" as keys. You can generate such a secret in your cluster using "kubectl create secret generic admin-password-secret-name --from-literal=password=admin --from-literal=currentPassword=admin"
```
leo-geoffroy-sonarsource, line 555: Let's wrap this line so it is not too long.
The general rule for SQ is 180

carminevassallo: yes, just wrapped both comments in standard and dce charts

leo-geoffroy-sonarsource (approved):

leo-geoffroy-sonarsource:
Tested, everything works fine

PR #252, opened 2022-08-26
DOCKER-81 Add support for monitoringPasscode passed as a secret and removal of in-clear-text livenessprobe httpheader

carminevassallo:
Currently, we cannot load the monitoringPasscode from an existing secret and we define it among liveness HTTP headers in clear text (DOCKER-65).

In this PR, we add the possibility to get this value from a secret and define a new liveness probe that masks the HTTP header.

### DIFF
--- a/charts/sonarqube-dce/values.yaml
+++ b/charts/sonarqube-dce/values.yaml
@@ -397,6 +397,9 @@ initFs:
 # a monitoring passcode needs to be defined in order to get reasonable probe results
 # not setting the monitoring passcode will result in a deployment that will never be ready
 monitoringPasscode: "define_it"
+# Alternatively, you can define the passcode loading it from an existing secret specifying the right key
+# monitoringPasscodeSecretName: "pass-secret-name"
+# monitoringPasscodeSecretKey: "pass-key"
 
 ## Environment variables to attach to the pods
 ##
--- a/charts/sonarqube-dce/CHANGELOG.md
+++ b/charts/sonarqube-dce/CHANGELOG.md
@@ -1,6 +1,9 @@
 # SonarQube Chart Changelog
 All changes to this chart will be documented in this file.
 
+## [4.2.0]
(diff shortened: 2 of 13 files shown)


leo-geoffroy-sonarsource:
I think we should not create the default secret `{{ template "sonarqube.fullname" . }}-monitoring-passcode` when we provision monitoringPasscodeSecretName and monitoringPasscodeSecretKey.
It is failing on my side since `.Values.monitoringPasscode` is not defined

leo-geoffroy-sonarsource (changes requested):
see comment above

carminevassallo:
> I think we should not create the default secret `{{ template "sonarqube.fullname" . }}-monitoring-passcode` when we provision monitoringPasscodeSecretName and monitoringPasscodeSecretKey. It is failing on my side since `.Values.monitoringPasscode` is not defined

Hi @leo-geoffroy-sonarsource,

thanks I just have a question. Isn't `monitoringPasscode` always defined in the values.yml? By default, we always have this value set to a dummy password (I'm referring to [this line](https://github.com/SonarSource/helm-chart-sonarqube/blob/2a3bed2f63fee521223a3bc1e085d9baf583d8bd/charts/sonarqube/values.yaml#L299)), so it should not happen that `monitoringPasscode` is unset. Am I right? Could it be that you were creating a secret having the same name and this was causing the error?

charts/sonarqube-dce/values.yaml
```diff
@@ -397,6 +397,9 @@ initFs:
 # a monitoring passcode needs to be defined in order to get reasonable probe results
 # not setting the monitoring passcode will result in a deployment that will never be ready
 monitoringPasscode: "define_it"
+# you can also define the passcode loading it from an existing secret specifying the right key
+# monitoringPasscodeSecretName: "passSecretName"
```
leo-geoffroy-sonarsource, line 401: maybe use a valid secret name as example

leo-geoffroy-sonarsource (approved):
LGTM

PR #256, opened 2022-09-07
Fix install plugins

HuKuToH:
Fix when plugins were error installed when using pvs due to incorrect directory mount


jacek-poreda-sonarsource:
Hello @HuKuToH ,

`lib/common` is not an appropriate place for installing anything regarding external plugins. Moreover, we plan to remove support for that directory in future versions see: https://sonarsource.atlassian.net/browse/DOCKER-93.

Thus I'm closing this PR.

PR #257, opened 2022-09-16
Hotfix - sonarSecretProperties conditionals

felipecafu95:
- [x] `ingressClass` key included in values.yaml for usage in Kubernetes cluster version >= 1.18
- [x] Conditionals at `charts/sonarqube-lts/templates/sonarqube-sts.yaml` updated from `and` to `or`, because only one key is necessary to "create" sonar.properties. In my use case, I used only the `sonarSecretProperties` for my all sonar configurations (like ldap, for example)

### DIFF
--- a/charts/sonarqube-lts/CHANGELOG.md
+++ b/charts/sonarqube-lts/CHANGELOG.md
@@ -1,6 +1,10 @@
 # SonarQube Chart Changelog
 All changes to this chart will be documented in this file.
 
+## [1.0.31]
+* IngressClass key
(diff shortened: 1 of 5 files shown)


carminevassallo:
Hey @felipecafu95,

thanks for your suggestion and contribution. We are gonna update the lts chart in a few weeks right after releasing the new sonarqube lts version. The new lts chart will include this feature. In the meantime, you might want to use the standard chart  :)

Kind Regards,
Carmine

PR #261, opened 2022-09-28
DRAFT: DOCKER-71-fix-wrong-condition-for-sonarSecretKey-in-helm-chart

jCOTINEAU:
- fix condition in helm templating for sonar.properties ConfigMap
- fix condition in helm templating for volumeMounts of this sonar.properties data

tests:
- templating locally with various .values file
- ct tests

### DIFF
--- a/charts/sonarqube-dce/templates/config.yaml
+++ b/charts/sonarqube-dce/templates/config.yaml
@@ -9,18 +9,15 @@ metadata:
     release: {{ .Release.Name }}
     heritage: {{ .Release.Service }}
 data:
-  {{- if and .Values.sonarSecretKey (not .Values.ApplicationNodes.sonarProperties) }}
-  sonar.properties: sonar.secretKeyPath={{ .Values.sonarqubeFolder }}/secret/sonar-secret.txt
-  {{- end }}
-  {{- if or .Values.ApplicationNodes.sonarProperties }}
+{{- if or .Values.sonarSecretKey .Values.ApplicationNodes.sonarProperties }}
   sonar.properties:
-  {{ range $key, $val := .Values.ApplicationNodes.sonarProperties }}
+  {{- range $key, $val := .Values.ApplicationNodes.sonarProperties }}
     {{ $key }}={{ $val }}
-  {{ end }}
   {{- end }}
-    {{- if and .Values.sonarSecretKey .Values.ApplicationNodes.sonarProperties }}
-      sonar.secretKeyPath={{ .Values.sonarqubeFolder }}/secret/sonar-secret.txt
-    {{- end }}
+  {{- if .Values.sonarSecretKey }}
+    sonar.secretKeyPath={{ .Values.sonarqubeFolder }}/secret/sonar-secret.txt
+  {{- end }}
+{{- end }}
 ---
 apiVersion: v1
 kind: ConfigMap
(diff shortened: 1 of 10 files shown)


charts/sonarqube-dce/templates/config.yaml
```diff
@@ -9,18 +9,15 @@ metadata:
...
     heritage: {{ .Release.Service }}
 data:
-  {{- if and .Values.sonarSecretKey (not .Values.ApplicationNodes.sonarProperties) }}
-  sonar.properties: sonar.secretKeyPath={{ .Values.sonarqubeFolder }}/secret/sonar-secret.txt
-  {{- end }}
-  {{- if or .Values.ApplicationNodes.sonarProperties }}
+{{- if or .Values.sonarSecretKey .Values.ApplicationNodes.sonarProperties }}
   sonar.properties:
-  {{ range $key, $val := .Values.ApplicationNodes.sonarProperties }}
+  {{- range $key, $val := .Values.ApplicationNodes.sonarProperties }}
```
leo-geoffroy-sonarsource, line 14: Is it working fine if .Values.ApplicationNodes.sonarPropertie is not specified at all?

charts/sonarqube/templates/deployment.yaml
```diff
@@ -162,7 +162,7 @@ spec:
               awk 1 /tmp/props/sonar.properties /tmp/props/secret.properties > /tmp/result/sonar.properties
             fi
           volumeMounts:
-          {{- if .Values.sonarProperties }}
```
leo-geoffroy-sonarsource, line 165: I think there is a condition missing here at line [146](https://github.com/SonarSource/helm-chart-sonarqube/pull/261/files#diff-d8492fb6e757528994fc1f4a506d435308d214cf5ed5903c9c9f6058ee14e12eL146)
The block will not be added if we only have a sonarSecretKey

charts/sonarqube/templates/sonarqube-sts.yaml
```diff
@@ -136,7 +136,7 @@ spec:
               awk 1 /tmp/props/sonar.properties /tmp/props/secret.properties > /tmp/result/sonar.properties
             fi
           volumeMounts:
-          {{- if .Values.sonarProperties }}
+          {{- if or .Values.sonarProperties .Values.sonarSecretKey ( not .Values.elasticsearch.bootstrapChecks) }}
```
leo-geoffroy-sonarsource, line 139: Maybe a suggestion here since we are repeating this condition, maybe we could add an helper in _helpers.tpl file such as isSonarConfigDefined

leo-geoffroy-sonarsource:
For testing this change, I would suggest the following:
- Use sonarSecretKey, sonarProperties, sonarSecretProperties, elasticsearch.bootstrapChecks independently then runs the helm chart and verify that in the pod, /opt/sonarqube/conf/sonar.properties and /opt/sonarqube/secret/sonar-secret.properties are defined as expected
- Activate multiple of these parameters at the same time and ensure the result in the same way

leo-geoffroy-sonarsource (changes requested):
Chart is failing + comments
Also missing a dash in commit message DOCKER-71 :)

charts/sonarqube-dce/templates/config.yaml
jCOTINEAU: Yes it output nothing if it is empty

charts/sonarqube/templates/deployment.yaml
jCOTINEAU: Good catch, changed things accordingly

jCOTINEAU:
> Chart is failing + comments Also missing a dash in commit message DOCKER-71 :)



> For testing this change, I would suggest the following:
> 
>     * Use sonarSecretKey, sonarProperties, sonarSecretProperties, elasticsearch.bootstrapChecks independently then runs the helm chart and verify that in the pod, /opt/sonarqube/conf/sonar.properties and /opt/sonarqube/secret/sonar-secret.properties are defined as expected
> 
>     * Activate multiple of these parameters at the same time and ensure the result in the same way

Thanks a lot for the good comments, passing the PR back in draft, until the tests are all done for the various parameters

leo-geoffroy-sonarsource:
Close the PR as the work is not finished

PR #263, opened 2022-10-10
Fix sonarProperties mount condition for DCE

kravvcu:
The DCE app pod is not getting it's sonarProperties volume mounted because it's (I'm guessing) using a deprecated .Values.sonarProperties key, rather then the nested one at .Values.ApplicationNodes.sonarProperties.

Example with installing object script (v4.0.7 helm chart version):

```
ApplicationNodes:
  plugins:
    install:
      - "https://objectscriptquality.com/sites/default/files/public/contents/plugin/release/sonar-objectscript-plugin-3.3.2.jar"

  sonarProperties:
    objectScriptQuality.license.serverUrl: https://www.objectscriptquality.com/api/v1/license
    objectScriptQuality.license.cache: /tmp/

# without this key defined only the `concat-properties` init container gets to mount the .Values.ApplicationNodes.sonarProperties
sonarProperties:
  dummy_key_needed_to_mount_sonar_properties_into_app_pod: true
```

### DIFF
--- a/charts/sonarqube-dce/CHANGELOG.md
+++ b/charts/sonarqube-dce/CHANGELOG.md
@@ -1,6 +1,9 @@
 # SonarQube Chart Changelog
 All changes to this chart will be documented in this file.
 
+## [4.0.8]
+* Fixed DCE chart to properly mount sonarProperties into the app pod
+
 ## [4.0.7]
 * Updated SonarQube to 9.6.1 for the DCE-chart application nodes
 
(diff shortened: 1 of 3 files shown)


carminevassallo:
Dear @kravvcu,

thanks for your investigation and contribution. This problem was fixed in a previous [PR](https://github.com/SonarSource/helm-chart-sonarqube/pull/249), which is already merged to master. The fix will be part of the next release 😉

Regards,
Carmine

kravvcu:
:facepalm: my bad, I was 100% sure I checked the latest master also.

PR #264, opened 2022-10-10
DOCKER-93 Fix the correct installation of plugins

carminevassallo:
Currently, our plugins are downloaded and installed in the `extensions/downloads` and `lib/common`. This is not what we state in our [documentation](https://docs.sonarqube.org/latest/setup/install-plugin/). In this PR, we mainly

- we fix the installation of plugins that must be in the `extensions/plugins` folder;
- we deprecate the usage of `plugins.lib`, which is a parameter that does not have a meaningful usage in the chart (`plugins.install` is all that we need).


charts/sonarqube-dce/README.md
```diff
@@ -201,8 +201,8 @@ The following table lists the configurable parameters of the Sonarqube chart and
 | `ApplicationNodes.prometheusExporter.httpsProxy` | HTTPS proxy for downloading JMX agent | `""` |
 | `ApplicationNodes.prometheusExporter.noProxy` | No proxy for downloading JMX agent | `""` |
 | `ApplicationNodes.prometheusExporter.securityContext` | Security context for downloading the jmx agent | see `values.yaml`|
-| `ApplicationNodes.plugins.install` | List of plugins to install | `[]` |
-| `ApplicationNodes.plugins.lib` | Plugins libray | `[]` |
+| `ApplicationNodes.plugins.install` | Link(s) to the plugin JARs to download and install | `[]` |
+| `ApplicationNodes.plugins.lib` | **DEPRECATED** Plugin libraries to download and install | `[]` |
```
leo-geoffroy-sonarsource, line 205: I'm not sure about the terminology here, we are saying it is deprecated, but at the same time we remove the behavior associated to it
I would say it is either we say it is deprecated and keep the behavior, then remove in the future, or remove entirely the possibility to set this value

charts/sonarqube-dce/templates/sonarqube-application.yaml
```diff
@@ -168,12 +168,9 @@ spec:
             "-e",
             "/tmp/scripts/install_plugins.sh"]
           volumeMounts:
-            - mountPath: {{ .Values.sonarqubeFolder }}/extensions/downloads
+            - mountPath: {{ .Values.sonarqubeFolder }}/extensions/plugins
```
leo-geoffroy-sonarsource, line 171: I'm thinking of a scenario where a plugin was installed before in download then after in plugin.
Is there a possibility that we have the plugin twice both in download and plugins? would it conflict at start?

charts/sonarqube-dce/README.md
carminevassallo: At the moment, this parameter is useless. It does not really contribute to installing the plugin. You can just assign a name to the jar using the `plugins.lib` parameter, which is not that useful in my opinion (why would need to rename the name of the plugin? 😃). The description is also not really clear, I've tried by myself and renaming the library downloaded from the `plugins.install` parameter is the only use you can do of `plugins.lib`. I was thinking of deprecating because people might still use `plugins.lib` but its functionality does not make sense in my opinion (I can keep the renaming capability during the deprecation period though)

charts/sonarqube-dce/templates/sonarqube-application.yaml
carminevassallo: I do not know, I can try and see what happens. Actually, existing installation in `extensions/downloads` should be "read" by the application, despite not being the official and documented way of installing plugins...

charts/sonarqube-dce/README.md
jCOTINEAU: I speak I: Even if it is not doing a lot and have a very low change of having a functional impact, i would prefer to write it deprecated, while keeping the feature for few release so that we are not removing something without notice.

charts/sonarqube-dce/templates/sonarqube-application.yaml
jCOTINEAU: I think there is no issue here, as we are only persisting that volume, after the pod restart the extensions/downloads will always be empty ?

leo-geoffroy-sonarsource: Side note from applicative point of view: at start, plugins are moved from download to plugins folder, so this + folder not being persisted, I think this is not possible to happen

charts/sonarqube-dce/README.md
carminevassallo: Generally, I agree with @jCOTINEAU. Even if the feature is not meaningful, it would be better to keep it entirely while being deprecated. However, we decided to remove the feature completely (and not deprecate it) as this is a misuse of the sonaqube-installation guidelines and it was never really supported by the product.

jCOTINEAU (approved):
LGTM

leo-geoffroy-sonarsource (approved):

PR #269, opened 2022-10-27
DOCKER-83 Define an initial set of dynamic compatibility tests

carminevassallo:
We need to define compatibility tests for our charts, that check whether a chart is compatible with a specific k8s version. In the scope of this PR, we focus on dynamic compatibility (e.g., how do our charts behave when we install them in a cluster of a certain k8s API version?).

We define one test suite for our standard and dce charts covering those basic three test cases. When we deploy our charts, assuming the usage of default values, we test:
- if the required (default) number of pods is running
- if all the running pods have the containersReady condition set to true
- if the sonarqube application is ready (apis are ready to serve new requests)


tests/sonarqube_standard_dynamic_test.go
```diff
@@ -0,0 +1,135 @@
...
+	helmChartPath := "../charts/" + chartName
+
+	// Setup the kubectl config and context. Here we choose to use the defaults, which is:
+	// - HOME/.kube/config for the kubectl config file
+	// - Current context of the kubectl config file
+	existingKubectlOptions := k8s.NewKubectlOptions("", "", "default")
+
+	// Define a new namespace
+	namespaceName := chartName + "-dynamic-test"
+	k8s.CreateNamespace(t, existingKubectlOptions, namespaceName)
```
jCOTINEAU, line 74: This fail if the ns already exist ( which can happen in case of previous run error etc )

On the setup of the test maybe we could ensure the ns does not exist ? it would also prevent leftovers ?

tests/sonarqube_standard_dynamic_test.go
```diff
@@ -0,0 +1,135 @@
...
+		ValuesFiles:    []string{helmChartPath+"/values.yaml"},
+		SetValues:      values,
+		KubectlOptions: kubectlOptions,
+	}
+
+	// Run RenderTemplate to render the template and capture the output.
+	output := helm.RenderTemplate(t, options, helmChartPath, chartName, []string{})
+	
+	// delete all resources after test execution
+	defer teardownTest(t, kubectlOptions, namespaceName)
```
leo-geoffroy-sonarsource, line 88: This should be put at the start of the method after L56 as it relates to the test execution workflow

leo-geoffroy-sonarsource:
Few indentation problems in the go files. try to run "go fmt" on the folder
As discussed, it would be good to decompose the tests in subfunctions for clarity and reusability

tests/sonarqube_standard_dynamic_test.go
leo-geoffroy-sonarsource: required kubectlOptions, so it cannot be put at the top

jCOTINEAU (commented):
Lgtm !

jCOTINEAU (approved):

tests/dynamic-compatibility-test/sonarqube_standard_dynamic_test.go
```diff
@@ -0,0 +1,200 @@
...
+	// Input values for our charts' tests
+	table := []struct {
+		name             string
+		chartName        string
+		expectedPods     int
+		values           map[string]string
+		valuesFilesPaths []string
+	}{
+		{"standard-chart-default-values-deployment", "sonarqube", 2, map[string]string{"tests.enabled": "false",
+			//"image.repository":           "sonarqube-arm64-v8", // DO NOT SET: This is only an experimental setting for M1 users
```
leo-geoffroy-sonarsource, line 61: can we remove the comments from the tests?
I think this should be documented elsewhere, and probably with the work on M1 this won't be needed anymore

leo-geoffroy-sonarsource (approved):

tests/dynamic-compatibility-test/sonarqube_standard_dynamic_test.go
carminevassallo: sure!

carminevassallo:
> LGTM, just the TODO to be removed, commented images and change jira ticket in commit

DOCKER-112 will address the todos that were removed from the code.

PR #276, opened 2022-12-01
added ability to set ingress class name if using networking.k8s.io/v1 API

krmichelos:
Ingresses as of `networking.k8s.io/v1` are supposed to use `ingressClassName` instead of the deprecated `kubernetes.io/ingress.class` annotation to control the ingress class of the ingress.  This chart supports creating ingresses in the networking.k8s.io/v1 API version but doesn't support the use of ingressClassName.  This pull requests adds the ability to conditionally set ingressClassName to allow users who use that exclusively to control ingress class use this chart.

- [X] Bump the Version number of the respected chart


jCOTINEAU:
Hello

PR #277, opened 2022-12-05
Update Readme.md contribution

jCOTINEAU:
It is not clear that our LTS chart is frozen, be more explicit

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,8 @@ Be aware that this forum is a community, so the standard pleasantries ("Hi", "Th
 Contributing
 ------------
 
+**Caution** our LTS chart is frozen as per our LTS policy, please do not ask or create modifications for that one.
+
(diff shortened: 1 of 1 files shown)


jCOTINEAU:
Hello @krmichelos, thanks a lot for taking the time to participate in the community.

For your question regarding the LTS policy, it is described [here](https://www.sonarsource.com/products/sonarqube/downloads/lts/)

If i try to summarize it, we will do changes to the LTS only if major blocker are detected. That is even more true now that we are close to releasing the next LTS.

May you excuse us for the inconvenience, we are aware this is not an optimal situation for our helm chart and will try our best to make the life of our user easier.

jCOTINEAU:
With the upcoming release of the new LTS, the release model will change, that comment in the README is irrelevant for now.

PR #285, opened 2022-12-23
Fix sonarqube dce search mounts

BitExodus:
Why

The generated yaml files had 3 elements declaring mount paths for the data, temp and logs folders when the init-fs init container was disabled, producing an invalid yaml with bad indentation and invalid for kubectl.

What

Move the "end" block to include the generation of the mount paths for data, temp and logs. The "if" block with the condition that checks if the init-fs container was closed too early in the previous version.

### DIFF
--- a/charts/sonarqube-dce/CHANGELOG.md
+++ b/charts/sonarqube-dce/CHANGELOG.md
@@ -1,6 +1,9 @@
 # SonarQube Chart Changelog
 All changes to this chart will be documented in this file.
 
+## [6.0.1]
+* Fix a bug in the generation of the Data Center Edition chart search StatefulSet.
+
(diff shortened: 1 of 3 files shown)


carminevassallo:
Hey @BitExodus,

thanks for spotting this issue and for proposing such a well-drafted PR! 

Only one thing to clarify: disabling the `init-fs` container won't cause the helm chart to produce badly formatted templates. The resulting k8s manifests are still valid. 

We will consider your change in one of the next hardening sprints. I have just written a small task about it ([SQHELM-110](https://sonarsource.atlassian.net/browse/SQHELM-110)) and we'll make sure to tackle it using the work you did on this feature branch.

For now, I'm closing this PR.

PR #286, opened 2022-12-29
SQHELM-1 Set a new maximum allowed size of the client request body

carminevassallo:
We do not recommend any maximum allowed payload sizes in our ingress configuration. This might end up in `Error 413` in a real scenario. Therefore, we add a new annotation in our chart with a default value set to 64m (after investigation).

### DIFF
--- a/charts/sonarqube-dce/README.md
+++ b/charts/sonarqube-dce/README.md
@@ -279,6 +279,7 @@ The following table lists the configurable parameters of the Sonarqube chart and
 | `ingress.hosts[0].servicePort` | Optional field to override the default servicePort of a path | `None` |
 | `ingress.tls` | Ingress secrets for TLS certificates | `[]` |
 | `ingress.ingressClassName` | Optional field to configure ingress class name | `None` |
+| `ingress.annotations` | Field to add extra annotations to the ingress | {`nginx.ingress.kubernetes.io/proxy-body-size=64m`} |
 | `ingress.annotations.nginx.ingress.kubernetes.io/proxy-body-size` | Field to set the maximum allowed size of the client request body  | `64m` |
 
 ### InitContainers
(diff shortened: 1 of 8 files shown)


charts/sonarqube-dce/README.md
```diff
@@ -279,7 +279,7 @@ The following table lists the configurable parameters of the Sonarqube chart and
 | `ingress.hosts[0].servicePort` | Optional field to override the default servicePort of a path | `None` |
 | `ingress.tls` | Ingress secrets for TLS certificates | `[]` |
 | `ingress.ingressClassName` | Optional field to configure ingress class name | `None` |
-| `ingress.annotations` | Optional field to add extra annotations to the ingress | `None` |
```
jCOTINEAU, line 282: Should we keep line 282 ? to let people know they can setup any annotation they like ?

Else just having the proxy-body-size is a bit misleading to me, wdyt ?

charts/sonarqube/README.md
```diff
@@ -194,7 +194,7 @@ The following table lists the configurable parameters of the Sonarqube chart and
 | `ingress.hosts[0].servicePort` | Optional field to override the default servicePort of a path | `None` |
 | `ingress.tls` | Ingress secrets for TLS certificates | `[]` |
 | `ingress.ingressClassName` | Optional field to configure ingress class name | `None` |
-| `ingress.annotations` | Optional field to add extra annotations to the ingress | `None` |
```
jCOTINEAU, line 197: same as for the DCE

charts/sonarqube-dce/README.md
carminevassallo: I was wondering the same, but what would I set as default value then? The whole ingress annotation?

jCOTINEAU: Hmm i dont know what who look best for user.

Maybe writing 'see below', pointing to the max_size default annotation that we are adding ?

carminevassallo: I have just restored the line with a new default value

jCOTINEAU (approved):
lgtm

PR #287, opened 2023-01-13
SQHELM-86 Update the supported kubernetes versions

carminevassallo:
We need to drop the support for EoL versions. This PR update the documentation, as direct consequence of the compatibility tests that we did before.


jCOTINEAU (approved):
lgtm

PR #279, opened 2022-12-12
feat(sonarqube/sonarqube-lts): allow setting priorityClassName for StatefulSets

Art3mK:
Hello 👋 

StatefulSets are using the same pod spec as deployments, and are supporting setting priorityClassNames too. This PR will allow user to use already existing value `priorityClassName` to change priority for sts pods too.

Thanks!

PR #299, opened 2023-02-08
SQHELM-109 allow setting priorityClassName for StatefulSets

jCOTINEAU:
* allow setting priorityClassName for StatefulSets

Co-authored-by: Artem Kajalainen <artem.kajalainen@virta.global>

This PR is based on the contribution made by Artem Kajalainenin the context of this [PR](https://github.com/SonarSource/helm-chart-sonarqube/pull/279)

### DIFF
--- a/charts/sonarqube-dce/CHANGELOG.md
+++ b/charts/sonarqube-dce/CHANGELOG.md
@@ -1,6 +1,9 @@
 # SonarQube Chart Changelog
 All changes to this chart will be documented in this file.
 
+## [8.1.0]
+* Allow setting priorityClassName for StatefulSets
+
 ## [8.0.2]
 * Fix the volume mounts for the init-fs container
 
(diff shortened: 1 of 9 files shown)


carminevassallo (approved):
The changes look good to me. Just some food for thought: shall we allow the users to assign search and app nodes to different priority classes? In all honesty, I think that SQ users just wanna make sure that all pods required for the application deployment are running, but more advanced users might also want to assign different priorities to the app nodes for example. Anyways, mine is just a comment for the future, I'm fine with merging this PR as is.

charts/sonarqube-dce/CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 # SonarQube Chart Changelog
 All changes to this chart will be documented in this file.
 
+## [8.0.2]
```
carminevassallo, line 4: We should increment the minor version for new features.

jCOTINEAU:
> The changes look good to me. Just some food for thought: shall we allow the users to assign search and app nodes to different priority classes? In all honesty, I think that SQ users just wanna make sure that all pods required for the application deployment are running, but more advanced users might also want to assign different priorities to the app nodes for example. Anyways, mine is just a comment for the future, I'm fine with merging this PR as is.

Totally agree, i dit not wanted to add complexity and new parameter so far as there is no direct needs, and i would assume people would like to set priority for the "SonarQube App" as a whole.

But definitely for advanced user and use case that could happen.

Agreeing on merging as is.

PR #273, opened 2022-11-06
Add configurable Prometheus PodMonitor resource

dpunkturban:
This PR adds the Prometheus PodMonitor as configurable resource, so you don't need to deploy the PodMonitor extra.

Please ensure your pull request adheres to the following guidelines:
- [x] explain your motives to contribute this change: what problem you are trying to fix, what improvement you are trying to make
- [x] Document your Changes in the `CHANGELOG.md` file of the respected chart as well as the `Chart.yaml`
- [x] Bump the Version number of the respected chart

PR #302, opened 2023-02-09
SQHELM-107 Add a configurable Prometheus PodMonitor resource

carminevassallo:
Co-authored-by: @dpunkturban

This PR is based on the contribution made by Dennis Urban in the context of [this PR](https://github.com/SonarSource/helm-chart-sonarqube/pull/273).


charts/sonarqube-dce/templates/prometheus-podmonitor.yaml
```diff
@@ -0,0 +1,55 @@
+{{- if and .Values.ApplicationNodes.prometheusExporter.enabled .Values.ApplicationNodes.prometheusExporter.podMonitor.enabled }}
+apiVersion: monitoring.coreos.com/v1
+kind: PodMonitor
+metadata:
+  name: {{ template "sonarqube.name" . }}
+  namespace: {{ .Values.ApplicationNodes.prometheusExporter.podMonitor.namespace | quote }}
+  labels:
+    app: {{ template "sonarqube.name" . }}
+spec:
+   {{- if  .Values.ApplicationNodes.prometheusExporter.podMonitor.jobLabel }}
```
jCOTINEAU, line 10: extra space after if ?

charts/sonarqube-dce/templates/prometheus-podmonitor.yaml
```diff
@@ -0,0 +1,55 @@
...
+    matchNames:
+    - {{ .Release.Namespace }}
+  selector:
+    matchLabels:
+      app: {{ template "sonarqube.name" . }}
+  podMetricsEndpoints:
+  - port: monitoring-ce
+    path: /
+    scheme: http
+    {{- if  .Values.ApplicationNodes.prometheusExporter.podMonitor.interval }}
```
jCOTINEAU, line 23: extra space after if

charts/sonarqube-dce/templates/prometheus-podmonitor.yaml
```diff
@@ -0,0 +1,55 @@
...
+  {{- end }}
+  namespaceSelector:
+    matchNames:
+    - {{ .Release.Namespace }}
+  selector:
+    matchLabels:
+      app: {{ template "sonarqube.name" . }}
+  podMetricsEndpoints:
+  - port: monitoring-ce
+    path: /
```
jCOTINEAU, line 21: Are we sure this '/' path will always be true ? 

Is there an SQ parameter that could modify that ? probably not that kind of stuff is handled by the reverse proxy i believe ?

charts/sonarqube-dce/templates/prometheus-podmonitor.yaml
```diff
@@ -0,0 +1,55 @@
...
+    {{- if  .Values.ApplicationNodes.prometheusExporter.podMonitor.interval }}
+    interval: {{ .Values.ApplicationNodes.prometheusExporter.podMonitor.interval }}
+    {{- end }}
+    {{- if .Values.ApplicationNodes.prometheusExporter.podMonitor.scrapeTimeout }}
+    scrapeTimeout: {{ .Values.ApplicationNodes.prometheusExporter.podMonitor.scrapeTimeout }}
+    {{- end }}
+  - port: monitoring-web
+    path: /
+    scheme: http
+    {{- if  .Values.ApplicationNodes.prometheusExporter.podMonitor.interval }}
```
jCOTINEAU, line 32: extra space after if

charts/sonarqube/templates/prometheus-podmonitor.yaml
```diff
@@ -0,0 +1,55 @@
+{{- if and .Values.prometheusExporter.enabled .Values.prometheusExporter.podMonitor.enabled }}
+apiVersion: monitoring.coreos.com/v1
+kind: PodMonitor
+metadata:
+  name: {{ template "sonarqube.name" . }}
+  namespace: {{ .Values.prometheusExporter.podMonitor.namespace | quote }}
+  labels:
+    app: {{ template "sonarqube.name" . }}
+spec:
+   {{- if  .Values.prometheusExporter.podMonitor.jobLabel }}
```
jCOTINEAU, line 10: same extra if as in dce

charts/sonarqube-dce/templates/prometheus-podmonitor.yaml
carminevassallo: I do not think there is a property that modifies it. But, while you were talking about this, I also realize that `monitoring-ce` and `monitoring-web` do not really expose prometheus metrics. Shall we keep just the real prometheus endpoint in the resource (i.e., `/api/monitoring/metrics`) as we specify in the [docs](https://docs.sonarqube.org/latest/instance-administration/monitoring/)?

jCOTINEAU: Yes there is something in the picture i still do not understand.

We have an init container that inject the prometheus exporter, that is no longer needed i believe, as it is now fully included in SQ no ? 

That's probably how it was done before, so yup i would agree we remove every mention to those things.

carminevassallo: It's indeed a tricky topic. As discussed, we both need to convert the JMX metrics and support the native prometheus metrics. For now, I was just decoupling the prometheus exporter and the prometheus PodMonitor. In my opinion, they are independent. Furthermore, I made sure that our podmonitor only supports the native prometheus endpoint. We need to discuss how to support such Prometheus exporter in the future...

charts/sonarqube/templates/secret.yaml
```diff
@@ -29,6 +29,22 @@ data:
   SONAR_WEB_SYSTEMPASSCODE: {{ .Values.monitoringPasscode | b64enc | quote }}
 {{- end }}
 ---
+{{- if and .Values.monitoringPasscode .Values.prometheusMonitoring.podMonitor.enabled}}
+apiVersion: v1
```
jCOTINEAU, line 33: Is that Secret a duplicate of the one above ? on line 18 to 30 ? Apart from the confition it looks like the same, and both would be applied when `if and .Values.monitoringPasscode .Values.prometheusMonitoring.podMonitor.enabled` condition is true ?

carminevassallo: Unfortunately, we need the same secret both in the release and the podmonitor namespaces. But such duplicate code can be refactored. I've just defined [a task](https://sonarsource.atlassian.net/browse/SQHELM-117) for it.

jCOTINEAU (approved):
lgtm.

PR #307, opened 2023-02-27
SQHELM-116 install_plugins.sh: fix plugins deletion

thomasgl-orange:
The `install_plugins.sh` script starts by trying to delete previously installed plugins, but this step silently fails in case there are several already installed, because of a shell script mistake (no plugin is deleted in this case). As a result, trying to update a plugin leads to having both the old and new versions present in the `extensions/plugins` folder, and thus SonarQube not starting.

Fix this by using `rm -f *` for cleaning up plugins, instead of broken `[ -e * ] && rm *`.

### DIFF
--- a/charts/sonarqube-dce/CHANGELOG.md
+++ b/charts/sonarqube-dce/CHANGELOG.md
@@ -1,6 +1,9 @@
 # SonarQube Chart Changelog
 All changes to this chart will be documented in this file.
 
+## [8.2.2]
+* Fix install_plugins.sh not deleting previously installed plugins
+
 ## [8.2.1]
 * Clarify doc for custom cacert secret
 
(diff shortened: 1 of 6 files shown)


thomasgl-orange:
Push-forced because this PR was now conflicting with recent changelog / chart version.

Also, I now came across [SQHELM-116](https://sonarsource.atlassian.net/browse/SQHELM-116), which I had not found the first time, so edited the PR title to mention it.

@carminevassallo, do you think you could review/merge this fix?  This issue (when several community plugins are installed) can be worked around with an extra init container to do the cleanup, but still, it would be nice not to have to :-)
Thanks.

carminevassallo:
Hey @thomasgl-orange, 

thanks for your contribution! Great job! Thanks also for linking the PR to the previously-raised issue and for spotting the bug, of course :)

This bug actually causes undesired behavior while installing the `sonarqube` chart only. The `sonarqube-dce` one is not affected as we don't mount the plugins on a persistent volume. However, as you highlighted in your PR the silent bug is also present in the dce edition's `install-plugins.sh` script and it makes sense to fix it everywhere.

We do have an issue on our CI for external contributions. To unblock this PR and let it get merged I've just drafted a [new PR](https://github.com/SonarSource/helm-chart-sonarqube/pull/311) (including a small edit on the change annotations in the charts). I close this, please follow the update there :)

PR #309, opened 2023-03-11
Add helper to flatten hashmap for sonarProperties

rhass:
This adds a helper to flatten sonarProperties dictionary so that the key is formatted properly. Without this helper the values for nested keys get improperly strigified as things like this:

```
sonar=map[governance:map[report:map[project:map[branch:map[frequency:Daily]]
```

This joins the strings with dots as expected and sets the values correctly.

### DIFF
--- a/charts/sonarqube-dce/CHANGELOG.md
+++ b/charts/sonarqube-dce/CHANGELOG.md
@@ -1,6 +1,9 @@
 # SonarQube Chart Changelog
 All changes to this chart will be documented in this file.
 
+## [8.3.2]
+* Add helper function to flatten dictionary keys and values to render properties correctly in properties file.
+
 ## [8.3.1]
 * Make `jvmOpts` and `jvmCeOpts` not override env vars and sonar properties
 
--- a/charts/sonarqube-dce/Chart.yaml
+++ b/charts/sonarqube-dce/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: sonarqube-dce
 description: SonarQube offers Code Quality and Code Security analysis for up to 27 languages. Find Bugs, Vulnerabilities, Security Hotspots and Code Smells throughout your workflow.
 type: application
-version: 8.3.1
+version: 8.3.2
 appVersion: 9.9.0
 keywords:
   - coverage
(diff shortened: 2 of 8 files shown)


rhass:
WRT: `Document your Changes in the CHANGELOG.md file of the respected chart as well as the Chart.yaml` -- Technically it's already documented, just makes it work as expected. I can update the docs though if y'all think it's needed though.

rhass:
Looks like there is a bug for when `sonarProperties` is null.

```template: sonarqube/templates/sonarqube-sts.yaml:36:28: executing "sonarqube/templates/sonarqube-sts.yaml" at <include (print $.Template.BasePath "/config.yaml") .>: error calling include: template: sonarqube/templates/config.yaml:12:17: executing "sonarqube/templates/config.yaml" at <.Values.sonarProperties.sonar>: nil pointer evaluating interface {}.sonar```

I will have to fix that when I get back in a couple weeks from a trip, but it's probably an easy fix. Suggestions welcome though. :smile:

rhass:
Eh... I fixed the bug really quick. :tada:

rhass:
Just updated the DCE chart, the CHANGELOG files, and bumped the versions.

Please let me know if there is anything else needed to get this merged. Thanks!

carminevassallo:
Dear @rhass,

first, thanks for opening the [community post](https://community.sonarsource.com/t/helm-chart-rendering-sonar-properties-key-names-incorrectly/85523/5) and then proposing a PR! It's truly awesome to have contributors that follow such a process. 

We added an insight to our product roadmap to evaluate the support of sonar.properties with nested keys. In any case, this would not mean that we stop supporting the current format, so we need to handle both in our charts. Given that this has not been prioritized yet, we won't work actively on the feature you are proposing but mostly act as reviewers. 

Having said that, do you still plan to work on it? We try to keep PRs not open for too long, so it would be great to know what are your plans on that. If you don't have time to work on it, you can still open the PR again when ready (or maybe the feature gets implemented in the meanwhile 😛).

rhass:
I am going to close this for now as it doesn't quite work as expected. There is an issue with the current implementation where it fails with multiple k/v pairs configured, which shouldn't happen. Unfortunately, I haven't the time to continue working on this.

PR #310, opened 2023-03-22
(Draft) SQHELM-116 install_plugins.sh: fix plugins deletion

carminevassallo:
**PR Draft needed to overcome the external PR issue**

The install_plugins.sh script starts by deleting previously installed plugins, but fails in case there are several already installed. As a result, updating a plugin leads to SonarQube not starting (because having two versions of the same plugin is a fatal error).

Fix this by using `rm -f *` instead of broken `[ -e * ] && rm *`.

- [ ] explain your motives to contribute this change: what problem you are trying to fix, what improvement you are trying to make
- [ ] Document your Changes in the `CHANGELOG.md` file of the respected chart as well as the `Chart.yaml`
- [ ] Bump the Version number of the respected chart

### DIFF
--- a/charts/sonarqube-dce/CHANGELOG.md
+++ b/charts/sonarqube-dce/CHANGELOG.md
@@ -1,6 +1,9 @@
 # SonarQube Chart Changelog
 All changes to this chart will be documented in this file.
 
+## [8.2.2]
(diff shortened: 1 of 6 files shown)


carminevassallo:
Drafter PR only to trigger the CI on the original coming from forked repos

PR #314, opened 2023-03-28
Comment out timeoutSeconds

bbodenmiller:

No need to set it since the value given is the default as noted at https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes


jCOTINEAU:
Hello @bbodenmiller, thanks a lot for taking the time to participate in the community.

We purposely chose to keep those default value explicitly defined, ensuring that any default changes will not impact users on that topic.

I am going to close that PR.

Nonetheless i would really thank you for looking for improving our product, please do not hesitate to keep giving us those feedback.

PR #322, opened 2023-04-21
SONAR-19119 fix wrong condition between initFS and concat-properties

jCOTINEAU:


charts/sonarqube-dce/templates/sonarqube-search.yaml
```diff
@@ -183,9 +183,9 @@ spec:
           volumeMounts:
             - name: init-fs
               mountPath: /tmp/scripts/
-{{- if .Values.searchNodes.persistence.mounts }}
-{{ toYaml .Values.searchNodes.persistence.mounts | indent 12 }}
-{{- end }}
+            {{- if .Values.searchNodes.persistence.mounts }}
+            {{- toYaml .Values.searchNodes.persistence.mounts | indent 12 }}
```
carminevassallo, line 187: Do we need the `-` here?

carminevassallo (approved):
LGTM, great catch! Just a silly question on the dash addition...

charts/sonarqube-dce/templates/sonarqube-search.yaml
jCOTINEAU: Good catch, yes it is needed but i forgot to change `indent` into `nindent` as per this [issue](https://github.com/helm/helm/issues/3854)

The goal was to have the `if/else` also indented for clarity, which requires to chomp space on the left

PR #324, opened 2023-04-25
Fix templates to workaround parse errors in Sonar-IaC

mstachniuk:


jCOTINEAU:
Hey @mstachniuk, regarding that PR we talked about, we will not have the time to work on it in the short term.

We are then closing the PR to clean up a bit and keeping the branch live so that we can reuse and extend your work later.

Thanks again for proposing those changes

PR #328, opened 2023-05-01
[NO-JIRA] Update cirrus config to skip promote task and run tests on nightly cron executed QA

matteo-mara-sonarsource:
The definition of the chart_testing task has been modified in order to run after each merge to master and during the nightly QA.

### DIFF
--- a/.cirrus.yaml
+++ b/.cirrus.yaml
@@ -1,4 +1,11 @@
-only_on_none_release_draft_template: &ONLY_ON_NONE_RELEASE_DRAFT_TEMPLATE
+env:
+### Shared variables
+  NIGHTLY_CRON: 'nightly-cron'
+
+except_nightly_cron: &EXCEPT_ON_NIGHTLY_CRON
+  only_if: $CIRRUS_CRON != $NIGHTLY_CRON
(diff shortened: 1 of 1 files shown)


.cirrus.yaml
```diff
@@ -35,7 +42,7 @@ clone_script_template: &CLONE_SCRIPT_TEMPLATE
     fi
 
 chart_testing_task:
-  only_if: $CIRRUS_PRERELEASE != "true" && $CIRRUS_BRANCH != "master"
+  <<: *ONLY_ON_NONE_RELEASE_DRAFT_TEMPLATE
```
jCOTINEAU, line 45: small typo here, should it be `non-release` instead of `nonE-release` ?

jCOTINEAU (approved):
added a comment on a small typo.

other than that lgtm, thanks a lot for the added value !

.cirrus.yaml
matteo-mara-sonarsource: Interestingly I was trying to understand what NONE_RELEASE meant, since I just reused what was already available in the file. I will update that!

PR #329, opened 2023-05-03
SONAR-18987 run test always, regardless of chart changes

jCOTINEAU:
`--all` will lint and test all charts, disabling changed charts detection and version increment checking.

Those checks are mostly causing merge conflict for user.

This change will allow us to release more easily on release-branches with the new release model.


leo-geoffroy-sonarsource (approved):
LGTM

PR #331, opened 2023-05-03
SONAR-18977 add postgresql and nginx value as deprecated

jCOTINEAU:


leo-geoffroy-sonarsource:
Are we sure we want to deprecate those features? Deprecate means we will remove them at some point in time, and I don't see any mention or decision on that in the original ticket

I see them more like the H2 database provided in sonarqube app, it is not for production use, but provided to quickly try it

jCOTINEAU:
Closing PR for now after some discussions.

current PR is not targeting directly the associated ticket.

Nonetheless it triggered a discussion that could super-seed it the initial ticket.

gathering some more information before moving on.

PR #333, opened 2023-05-09
BUILD-2953  Adapt helm-chart-sonarqube (replace set-output deprecated)

sebastienvermeille:
# BUILD-2953  Adapt helm-chart-sonarqube (replace set-output deprecated)

## Changes

* Replace deprecated `set-output` with the [recommendations of GitHub](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/)
   That way it can work after 31st of May


carminevassallo (approved):
@sebastienvermeille I've checked the GitHub recommendations you just linked and your change makes sense to me! 😃

PR #343, opened 2023-06-26
Add PDB for sonarqube helm chart

SamuelBagattin:

Problem : Cluster autoscaler or Karpenter consider that the Sonarqube pod can be evicted (in case of an underutilized node for example), which can cause application downtime as this chart does not deploy an highly available sonarqube.

Solution : allow to setup a PodDisruptionBudget to block the eviction of the pod.

### DIFF
--- a/charts/sonarqube/CHANGELOG.md
+++ b/charts/sonarqube/CHANGELOG.md
@@ -1,6 +1,9 @@
 # SonarQube Chart Changelog
 All changes to this chart will be documented in this file.
 
+## [10.1.1]
+* Add PodDisruptionBudget
+
 ## [10.1.0]
 * Update SonarQube to 10.1.0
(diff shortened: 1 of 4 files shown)


carminevassallo:
Hey @SamuelBagattin,

thanks for your proposed contribution! 

We discussed this internally a bit and we are not gonna accept the change this time. We do use PDB on the `sonarqube-dce` chart, that's the SQ version where we focus more on those resiliency aspects. This chart/edition also enables you to scale SQ across multiple replicas.

Nevertheless, you might want to define a PriorityClass in your cluster and use the `priorityClassName` parameter to "protect" SonarQube.

Having said that, we are looking forward to your next contribution, and feel free to post in our [forum](https://community.sonarsource.com/c/help/sq/10) too.

PR #347, opened 2023-07-03
Add a possibility to adjust the Helm hook phase for admin job

dreamblack86:
The exact value of this annotation is configured using the Helm values file (Values). The default value is "post-install", but it can be overridden via .Values.adminJobHelmHook.

We need this because we roll out the helmet map in a CI/CD pipeline and always use Helm upgrade here. If we want to change the admin password now, we can't implement that with the "post install". The Helm release would then have to be deleted and reinstalled. But if we change the hook to "post-upgrade", the job will be created even though the version was already installed.

If the password was set in the past, this is not a problem for the job, as it simply would not be authorized (HTTP 401) to change the password. However, the curl command will still exit with "exit 0". 

However, if you want to change the password, you must always adjust "currentAdminPassword" and "adminPassword" so that the job is actually authorized to change the password.

### DIFF
--- a/charts/sonarqube-dce/CHANGELOG.md
+++ b/charts/sonarqube-dce/CHANGELOG.md
@@ -1,6 +1,10 @@
 # SonarQube Chart Changelog
 All changes to this chart will be documented in this file.
 
+## [10.3.0]
+* Update SonarQube to 10.3.0
+* Add a possibility to adjust the Helm hook phase for admin job
+
 ## [10.2.0]
 * Update Chart's version to 10.2.0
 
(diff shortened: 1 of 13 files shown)


carminevassallo:
Dear @dreamblack86,

Thanks for your contribution, this is such a great suggested improvement!

I was reviewing your use case and I agree that we should make this chart's hook phase configurable, to handle the case of upgrades. 

However, before accepting your PR as is, I would like not to make the phase fully configurable (for example, I'd like to avoid the hook being wrongly used with `pre-install`). What I mean is that we should set a constraint in our schema validation (`values.schema.json`) to make sure that the only allowed values are `post-install` and `post-upgrade` (cannot think about other valid cases). 

I've drafted a ticket for it ([SONAR-20474](https://sonarsource.atlassian.net/browse/SONAR-20474)) and tackle this topic in one of the next hardening sprints.

PR #349, opened 2023-07-20
SONAR-19950 Updates cURL image to 8.2.0

davividal:
Updates the cURL image we are using from 7.76.1 to 8.2.0

### DIFF
--- a/charts/sonarqube-dce/CHANGELOG.md
+++ b/charts/sonarqube-dce/CHANGELOG.md
@@ -3,6 +3,7 @@ All changes to this chart will be documented in this file.
 
 ## [10.2.0]
 * Update Chart's version to 10.2.0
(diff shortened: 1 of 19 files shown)


carminevassallo (changes requested):
Hey @davividal, apart from fixing the changelog as we discussed, please consider the following changes to your PR :)

- For each chart, we should update the README, as we have now a new default value for `curlContainerImage`
- Sometimes we pinpoint the version, sometimes not (I've also updated the context in the task description). Could you please update every occurrence of the curlimages/curl image to use 8.1.2 (or even 8.2.0)? In `change-admin-password-hook.yml`, we use "curlimages/curl:latest" as default if the value is not specified.

carminevassallo (approved):
LGTM!

PR #351, opened 2023-07-28
SONAR-20054 Set the minimum values for PodDisruptionBudget

carminevassallo:


charts/sonarqube-dce/values.yaml
```diff
@@ -145,7 +145,7 @@ ApplicationNodes:
 
   # Pod Distribution Budget for search nodes
```
jCOTINEAU, line 146: Typo in comment, should mention APP instead of Search.

jCOTINEAU (approved):
lgtm.

charts/sonarqube-dce/values.yaml
carminevassallo: good catch!

PR #352, opened 2023-08-03
SONAR-19951 Fixes podMonitoringEndpoints to use the correct endpoint

davividal:


carminevassallo (commented):
Hey @davividal, 

I think that this is not a bug. To reach the endpoint `/api/monitoring/metrics` we must rely on the exposed `http` port. The metrics are also made ready to be collected by Prometheus (look at [the documentation](https://docs.sonarsource.com/sonarqube/latest/instance-administration/monitoring/#prometheus-monitoring)).

Let me know what you think...

PR #362, opened 2023-09-07
[NO JIRA] Remove strict check on version bumps

carminevassallo:
We adopt a new release policy where every change does not require a bump on the chart's version. This PR removes an old check and align the LTS policy with what we follow on the main development stream.


carminevassallo:
> that means we start using the "follow sq version" also on LTS branch ?

Nope, not really. But in terms of patch numbers, it would be nice to keep some alignments to avoid generating confusion. For example, if we release SQ LTS 9.9.2, it would create less ambiguity if the patch number is the same on the chart (e.g., 7.0.2). WDYT?

jCOTINEAU (approved):
Considering the patch frequency on LTS i agree it would bring more clarity. lgtm.

PR #363, opened 2023-09-12
Add comment to nginx annotation

daniel-maganto:

Add comment for ingress nginx annotation because it is not mandatory if you are not using nginx ingress controller and create fails for other kind of ingressClassName.


daniel-maganto:
@carminevassallo could you take a look?

carminevassallo:
Dear @daniel-maganto,

Thanks a lot for your contribution and interest in our chart.

I agree with you, we have a nginx-specific annotation when defining the Ingress resource. However, we can't simply remove the default annotation, as this would cause issues for people who want to try our chart relying on the ingress-nginx controller. Specifically, they might face issues when uploading "big" reports to SonarQube.

Therefore, instead of accepting the change right away, I just drafted an improvement task [SONAR-20470](https://sonarsource.atlassian.net/browse/SONAR-20470): we'll refactor that part of the chart to make it the annotation not the default when nginx is not used. This will be tackled in one of our next hardening sprints.

PR #366, opened 2023-09-20
SONAR-20470 Set the nginx-proxy-body size only when nginx is enabled

carminevassallo:


jCOTINEAU (commented):
Great improvment overhall ! 

Only question/drawback I have is the situation where a customer was using its own ingress controller, without overriding annotations, so just having `ingress: enabled: true`

Now with the new value file and the condition, this annotation will be removed, which can be a small drawback.

Step to reproduce:

From master
```
ingress:
  enabled: true
```
`helm template --dry-run --debug . | grep body-size` return true as we have the explicit default.

While on your branch the same command return false because the default is no longer added as it is removed from the annotations value.

Should we add this default in any case ? so keeping your logic but removing the condition `{{- if or .Values.ingress.annotations .Values.nginx.enabled}}` because if no annotations exist, the range will do nothing and the `{{- if and .Values.nginx.enabled (not (hasKey (.Values.ingress.annotations) "nginx.ingress.kubernetes.io/proxy-body-size"))` will always write down the default ? 

wdyt ?

carminevassallo:
> While on your branch the same command return false because the default is no longer added as it is removed from the annotations value.

Yes, but this is expected. When the nginx is disabled and only ingress is enabled, I suggest that the body size is not set. I understand that this is a drawback for people using nginx on their own (without installing it through our chart) but we can't really detect when nginx is installed "separately" from our chart. Furthermore, if we leave it as default value, people relying on a different ingress technology will have a default annotation that does not make sense in their case (and this was the WHY for this PR #363).

WDYT, @jCOTINEAU?

jCOTINEAU (approved):
LGTM !

daniel-maganto:
Many thanks for the implementation!

carminevassallo:
> Many thanks for the implementation!

Thank you again @daniel-maganto for suggesting it!

PR #367, opened 2023-09-21
fix label issue with digest images

jan-kantert:
When tring to set a digest to the image tag the helm chart fails with the following error:

`Upgrade "sonarqube" failed: cannot patch "sonarqube-sonarqube" with kind Deployment: Deployment.apps "sonarqube-sonarqube" is invalid: [metadata.labels: Invalid value: "lts-community@sha256:3596d14feb065a31ce84cef60cc3ecfb7b47233ef860fd85c0d4e465f676c9f7": must be no more than 63 characters, metadata.labels: Invalid value: "lts-community@sha256:3596d14feb065a31ce84cef60cc3ecfb7b47233ef860fd85c0d4e465f676c9f7": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')]`

This change escapes all non word characters (especially @ in this case) and truncates the label to 63 characters.

### DIFF
--- a/charts/sonarqube/CHANGELOG.md
+++ b/charts/sonarqube/CHANGELOG.md
@@ -4,6 +4,7 @@ All changes to this chart will be documented in this file.
 ## [10.3.0]
 * Update Chart's version to 10.3.0
 * Update default images to the latest versions
+* Fix crash when adding a digests to the image tag
 
 ## [10.2.0]
 * Update SonarQube to 10.2.0
--- a/charts/sonarqube/templates/deployment.yaml
+++ b/charts/sonarqube/templates/deployment.yaml
@@ -13,7 +13,7 @@ metadata:
     app.kubernetes.io/managed-by: {{ .Release.Service }}
     app.kubernetes.io/part-of: sonarqube
     app.kubernetes.io/component: {{ template "sonarqube.fullname" . }}
-    app.kubernetes.io/version: {{ tpl .Values.image.tag . | quote }}
+    app.kubernetes.io/version: {{ tpl .Values.image.tag . | regexReplaceAll "\\W+" .Release.Name "_" | trunc 63 | trimSuffix "-" | quote }}
 spec:
   replicas: {{ .Values.replicaCount }}
   selector:
--- a/charts/sonarqube/templates/deployment.yaml
+++ b/charts/sonarqube/templates/deployment.yaml
@@ -13,7 +13,7 @@ metadata:
     app.kubernetes.io/managed-by: {{ .Release.Service }}
     app.kubernetes.io/part-of: sonarqube
     app.kubernetes.io/component: {{ template "sonarqube.fullname" . }}
-    app.kubernetes.io/version: {{ tpl .Values.image.tag . | regexReplaceAll "\\W+" .Release.Name "_" | trunc 63 | trimSuffix "-" | quote }}
+    app.kubernetes.io/version:  {{ regexReplaceAll "\\W+" (tpl .Values.image.tag .) "_" | trunc 63 | trimSuffix "-" | quote }}
 spec:
   replicas: {{ .Values.replicaCount }}
   selector:


jan-kantert:
@jCOTINEAU is there anything I should do to move this forward?

jCOTINEAU:
Hello @jan-kantert Sorry for the small delay, i will trigger the tests on the CI while testing manually.

On my side, when checking out your branch and running
`helm template  -n sonarqube --set deploymentType=Deployment .`

i get the following error:
`Error: template: sonarqube/templates/deployment.yaml:16:60: executing "sonarqube/templates/deployment.yaml" at <regexReplaceAll>: wrong number of args for regexReplaceAll: want 3 got 4`
I am guessing it is because the initial | inject the version onto the regexReplaceAll which end up having the version as an extra args ?

Could it be something like that instead ? 
`{{ regexReplaceAll "\\W+" (tpl .Values.image.tag .) "_" | trunc 63 | trimSuffix "-" | quote }}`

jan-kantert:
> Hello @jan-kantert Sorry for the small delay, i will trigger the tests on the CI while testing manually.
> 
> On my side, when checking out your branch and running `helm template -n sonarqube --set deploymentType=Deployment .`
> 
> i get the following error: `Error: template: sonarqube/templates/deployment.yaml:16:60: executing "sonarqube/templates/deployment.yaml" at <regexReplaceAll>: wrong number of args for regexReplaceAll: want 3 got 4` I am guessing it is because the initial | inject the version onto the regexReplaceAll which end up having the version as an extra args ?
> 
> Could it be something like that instead ? `{{ regexReplaceAll "\\W+" (tpl .Values.image.tag .) "_" | trunc 63 | trimSuffix "-" | quote }}`

Yes that looks better. I will apply that.

> While i understand the reasoning why the chart should be sure to produce valid yaml, triming all field would be cumbersome, is it something that could be fixed when passing the problematic value ?

I am not completely sure if we actually need this label at all. Might be useful in some cases but it also just mirrors the image field of the container.

jCOTINEAU:
In that case and if you are fine with it, if we do not see a very strong and clear need for adding that feature considering what we said, i will close the PR

jan-kantert:
> In that case and if you are fine with it, if we do not see a very strong and clear need for adding that feature considering what we said, i will close the PR

@jCOTINEAU  Its still broken the way it is. It is not possibe to use LTS tags. Should we fork the chart locally instead?

Solution should be either:
- Make sure the label is not getting too long
- or: Remove the label

Doing neither does not solve it.

jCOTINEAU:
I might have overlooked the issue and not understood correctly the issue.

Can you give us the sample value file with every sensitive stuff removed? 
Does the issue occurs because you use `image: ` with the image `sha` and not the tag ? Was this working before or is this a first start ?

jan-kantert:
> I might have overlooked and not understood correctly the issue, may you excuse me
> 
> Can you give us the sample value file with every sensitive stuff removed? Does the issue occurs because you use `image: ` with the image `sha` and not the tag ? Was this working before or is this a first start ?

This is working without a sha. However, we would like to set explicit image versions so we added a sha. Especially since there is only one `lts` tag (and no 9-lts, 10-lts etc). Otherwise, we might end up installing a major update over night.

jCOTINEAU:
That is indeed a very fair use case considering we do not provide lts version specific tags.

Considering we have a lot of `app.kubernetes.io/version:` iterations, could you factorize everything under a template like in `charts/sonarqube/templates/_helpers.tpl` and propagate also on the dce chart ?

Huge thanks for you contributions.

jCOTINEAU:
Hello @jan-kantert, huge thanks again for contributing to our helm chart.

As it has been quite a while, I will close the PR and create an hardening [ticket on our backlog](https://sonarsource.atlassian.net/jira/software/c/projects/SONAR/boards/350?selectedIssue=SONAR-21534)

When we will work on this task, we will reuse your work and acknowledge you as a contributor.

PR #370, opened 2023-09-27
SONAR-20543 fix docker rate limit by adding pullsecret login

jCOTINEAU:
fix docker rate limit by adding pullsecret login
NO-JIRA fix typo in cirrus yaml


carminevassallo (approved):
Approved, and it does not look as bad as you were thinking before :P

PR #371, opened 2023-10-03
Update default memory resources to provide for the Enterprise Edition

sylvain-combe-sonarsource:
The same Helm Chart is used to deploy all single pod SonarQube editions, and while Community and Developer run well on a 4 GB RAM pod, it's not the case for the Enterprise Edition which requires 6 GB.

- [ ] explain your motives to contribute this change: what problem you are trying to fix, what improvement you are trying to make
- [ ] Document your Changes in the `CHANGELOG.md` file of the respected chart as well as the `Chart.yaml`

### DIFF
--- a/charts/sonarqube/values.yaml
+++ b/charts/sonarqube/values.yaml
@@ -332,16 +332,19 @@ monitoringPasscode: "define_it"
 # Set annotations for pods
 annotations: {}
 
-## We usually don't make specific ressource recommandations, as they are heavily dependend on
-## The usage of SonarQube and the surrounding infrastructure.
-## Adjust these values to your needs, but make sure that the memory limit is never under 4 GB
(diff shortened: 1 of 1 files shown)


carminevassallo:
Hey @sylvain-combe-sonarsource,

thanks for proposing the PR! We would like to tackle the topic in one of the next hardening sprints, including this [ticket](https://sonarsource.atlassian.net/browse/SONAR-20656) into the scope.

The reason why we decided not to accept your suggestions right away is that we want to quickly assess the impact of users that rely on default values and adopting an autoscaling process...(for example, while dogfooding our own node autoscaling process we were experiencing issues related to such memory limits). I'm closing this PR for now, as we will pick the topic again when tackling the ticket above...

PR #372, opened 2023-10-16
SONAR-20766 Separate annotations and labels for app and search pods

carminevassallo:


jCOTINEAU (approved):
lgtm! small but really valuable improvment !

PR #374, opened 2023-10-23
SONAR-20810 change default security context value to comply with pod …

jCOTINEAU:
security standard

### DIFF
--- a/charts/sonarqube-dce/values.yaml
+++ b/charts/sonarqube-dce/values.yaml
@@ -69,7 +69,15 @@ searchNodes:
   # Set security context for sonarqube search container
   containerSecurityContext:
     # Sonarqube dockerfile creates sonarqube user as UID and GID 1000
+    # Those default are used to match pod security standard restricted as least privileged approach
+    allowPrivilegeEscalation: false
+    runAsNonRoot: true
     runAsUser: 1000
+    runAsGroup: 1000
+    seccompProfile:
+      type: RuntimeDefault
+    capabilities:
+      drop: ["ALL"]
 
   # Probe threshold configuration for the search nodes
   readinessProbe:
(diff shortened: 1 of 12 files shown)


charts/sonarqube-dce/values.yaml
```diff
@@ -69,7 +69,15 @@ searchNodes:
   # Set security context for sonarqube search container
   containerSecurityContext:
     # Sonarqube dockerfile creates sonarqube user as UID and GID 1000
+    # Those default are used to match pod security standard restricted as least privileged approach
```
davividal, line 72: Can't we use YAML references to avoid duplication?

I'm particularly not very fond of it, but...

```yaml
defaultContainterSecurityContext: & defaultContainterSecurityContext
  allowPrivilegeEscalation: false
  runAsNonRoot: true
  runAsUser: 1000
  runAsGroup: 1000
  seccompProfile:
    type: RuntimeDefault
  capabilities:
    drop: ["ALL"]
...
  containerSecurityContext:
    <<: *defaultContainterSecurityContext
```

davividal (commented):
@jCOTINEAU, what do you think about my comment?
Other than that, it looks good to me! Amazing work!

davividal (changes requested):
Can you double-check why the tests failed, please?

charts/sonarqube-dce/values.yaml
jCOTINEAU: Ha, that's a very good question. i think we should discuss that with Carmine, when we define how we want to structure and architecture the helm chart.

In the meantime, not found of it either, will keep as is even if quite verbose.

jCOTINEAU:
Tests was not passing due to a lack of proper capability on the initFs that i missed, all green, thanks a lot !

jCOTINEAU (commented):
Code has been changed to fix the lack of permission, code request tackled

PR #379, opened 2023-10-30
SONAR-20657 - Reduces k8s timeout to be less than Cirrus'

davividal:


carminevassallo (changes requested):
Nice try to gain more knowledge about this timeout 👏 

I saw that their a `dump` command in `kubectl` to get even more info for debugging purposes while the general `cluster-info` should be not that informative (however, I think it's still good to have).

Therefore, for the time being, I'm proposing to have them both using something like

```
# Timeout reached, Kind cluster is not available
echo "Timeout reached, Kind cluster is not available"
kubectl cluster-info --context kind-kind 2>&1
echo "-------------------------"
echo "Detailed logs:"
kubectl cluster-info dump --context kind-kind 2>&1
```

WDYT?

carminevassallo (approved):

PR #386, opened 2023-11-09
SONAR-20657 change wait for kind behavior

jCOTINEAU:
Add a retry mechanism at creation time, this should prevent flakyness


davividal (approved):
LGTM!

PR #387, opened 2023-11-09
SQDEVOPS-309 update nginx dependency

jCOTINEAU:


charts/sonarqube-dce/Chart.yaml
```diff
@@ -70,6 +70,6 @@ dependencies:
     repository: https://raw.githubusercontent.com/bitnami/charts/pre-2022/bitnami
     condition: postgresql.enabled
   - name: ingress-nginx
-    version: 4.7.1
+    version: 4.8.3
```
carminevassallo, line 73: Hey @jCOTINEAU,

shall we log this update in the change list? I think we haven't always done this in the past, but we should. WDYT?

carminevassallo:
A minor comment on the commit message. Shall we adopt the following structure for this and future dependency updates? "Update X to version Y"? If yes, we might add a note in the helm dev guidelines...

charts/sonarqube-dce/Chart.yaml
jCOTINEAU: that is a very good catch doing it now

carminevassallo (approved):
LGTM!

PR #397, opened 2023-11-20
Fix empty ingress annotations

zoltanlaczko:
Change ingress annotations from null instead of empty dict

### DIFF
--- a/charts/sonarqube/values.yaml
+++ b/charts/sonarqube/values.yaml
@@ -100,7 +100,7 @@ ingress:
       # servicePort: somePort
       # the pathType can be one of the following values: Exact|Prefix|ImplementationSpecific(default)
       # pathType: ImplementationSpecific
-  annotations:
+  annotations: {}
   # kubernetes.io/tls-acme: "true"
 
   # Set the ingressClassName on the ingress record


davividal (changes requested):
Hi @zoltanlaczko. Thank you for taking the time to submit this patch.

Could you replicate your fix to our `sonarqube-dce` chart and update the Changelogs at `Chart.yaml` and `CHANGELOG.md` in both charts?

Could you also please clarify your use case or what is the issue you are facing? I just tested the chart with the following `values.yaml`:

```yaml
ingress:
  enabled: true

nginx:
  enabled: true
```

And this is my resulting ingress controller:

```
Name:             sonarqube-sonarqube
Labels:           app=sonarqube
                  app.kubernetes.io/managed-by=Helm
                  chart=sonarqube-10.4.0
                  heritage=Helm
                  release=sonarqube
Namespace:        sonarqube
Address:
Ingress Class:    <none>
Default backend:  <default>
Rules:
  Host                    Path  Backends
  ----                    ----  --------
  sonarqube.your-org.com
                          /   sonarqube-sonarqube:9000 ()
Annotations:              meta.helm.sh/release-name: sonarqube
                          meta.helm.sh/release-namespace: sonarqube
                          nginx.ingress.kubernetes.io/proxy-body-size: 64m
Events:                   <none>
```

Which conforms to our [README](https://github.com/SonarSource/helm-chart-sonarqube/blob/master/charts/sonarqube/README.md#ingress-1) and our [ingress template](https://github.com/SonarSource/helm-chart-sonarqube/blob/master/charts/sonarqube/templates/ingress.yaml#L16-L24). Am I missing something here?

zoltanlaczko:
The nginx-ingress is already installed on my cluster, so it will fail if I enable it.

My issue is the [rancher manager](https://ranchermanager.docs.rancher.com/) way of releasing helm charts. It merges the chart`s values.yaml and the values what you set during the release.

The `nginx.annotations` is `null` and `null` merged with a `{}` (dict) will result a `null`.

Not sure but I think this can be reproduced also if you release without annotations and upgrade with annotations.

charts/sonarqube-dce/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: sonarqube-dce
 description: SonarQube is a self-managed, automatic code review tool that systematically helps you deliver clean code. As a core element of our Sonar solution, SonarQube integrates into your existing workflow and detects issues in your code to help you perform continuous code inspections of your projects. The tool analyses 30+ different programming languages and integrates into your CI pipeline and DevOps platform to ensure that your code meets high-quality standards.
 type: application
-version: 10.4.0
+version: 10.4.1
```
davividal, line 5: Can you add a changelog also under the `artifacthub.io/changes` [annotation](https://artifacthub.io/docs/topics/annotations/helm/#supported-annotations), please?

charts/sonarqube/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: sonarqube
 description: SonarQube is a self-managed, automatic code review tool that systematically helps you deliver clean code. As a core element of our Sonar solution, SonarQube integrates into your existing workflow and detects issues in your code to help you perform continuous code inspections of your projects. The tool analyses 30+ different programming languages and integrates into your CI pipeline and DevOps platform to ensure that your code meets high-quality standards.
 type: application
-version: 10.4.0
+version: 10.4.1
```
davividal, line 5: Same here.

davividal (changes requested):
Hi @zoltanlaczko. I appreciate your clarification. It makes sense.

I need these final two changes before I can approve your PR.

charts/sonarqube-dce/Chart.yaml
zoltanlaczko: done

charts/sonarqube/Chart.yaml
zoltanlaczko: done

charts/sonarqube-dce/CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 # SonarQube Chart Changelog
 All changes to this chart will be documented in this file.
 
+## [10.4.1]
+* Update Chart's version to 10.4.1
+* Fix empty ingress annotations in values
```
davividal, line 6: Could you please move the `Fix empty ingress annotations in values` to the `10.4.0` version?

I failed to notice this before, but according to our release policy, this fix must be under the `10.4.0` version. The same applies to the other changes in this PR.

davividal (changes requested):
Hi @zoltanlaczko. Thank you again for your contribution.

Can you please keep everything under the `10.4.0` version so it will comply with our release policy? :)
I should have noticed this before; I'm sorry.

davividal (approved):
LGTM!

zoltanlaczko:
@davividal @jCOTINEAU 

is there anything else I can do to merge?

jCOTINEAU:
hello @zoltanlaczko, i created this [PR](https://github.com/SonarSource/helm-chart-sonarqube/pull/406) that will trigger tests on our side.

In the meantime is see there are small conflicts with your PR, could you fix them?

PR #399, opened 2023-11-30
Run the initSysctl initcontainer as root

Mikaciu:
Kubernetes version: v1.26.6
When not using the root user, the initSysctl throws an error: 
```sysctl: permission denied on key "vm.max_map_count", ignoring```

After this, elasticSearch refuses to start:
```bootstrap check failure [1] of [1]: max virtual memory areas vm.max_map_count [65530] is too low, increase to at least [262144]```

### DIFF
--- a/charts/sonarqube-dce/CHANGELOG.md
+++ b/charts/sonarqube-dce/CHANGELOG.md
@@ -4,6 +4,7 @@ All changes to this chart will be documented in this file.
 ## [10.4.0]
 * Update Chart's version to 10.4.0
 * Improve the description of deprecated `ApplicationNodes.jvmOpts` and `ApplicationNodes.jvmCeOpts` values
+* Run the initSysctl init-container as root to prevent 'permission denied' issues
 
 ## [10.3.0]
 * Upgrade SonarQube to 10.3.0
@@ -239,7 +240,7 @@ All changes to this chart will be documented in this file.
 * added link to community support forum
 
 ## [0.1.6]
-* fixed wrong scc user reference if name was explicitly set 
+* fixed wrong scc user reference if name was explicitly set
 
 ## [0.1.5]
 * fixed serviceaccount logic
--- a/charts/sonarqube-dce/Chart.yaml
+++ b/charts/sonarqube-dce/Chart.yaml
@@ -31,6 +31,8 @@ annotations:
       description: "Update Chart's version to 10.4.0"
     - kind: fixed
       description: "Improve the description of deprecated 'ApplicationNodes.jvmOpts' and 'ApplicationNodes.jvmCeOpts' values"
+    - kind: fixed
+      description: "Run the initSysctl init-container as root to prevent 'permission denied' issues"
   artifacthub.io/links: |
     - name: support
       url: https://community.sonarsource.com/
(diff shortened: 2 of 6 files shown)


Mikaciu:
> Hello @Mikaciu, thanks a lot for taking the time to participate in the community.
> 
> We are investigating why it has not triggered the same failure on our EKS clusters, which is quite surprising.
> 
> Could you also move the changes to 10.4.0 instead of 10.4.1 ? Inside the helm chart we try to follow the SonarQube versioning in order to have a clear 1-1 mapping of the versions.
> 
> Also if you can pack up those changes in the sonarqube-dce would be super great.
> 
> Thanks again, that is a really good contribution.

Thanks, glad to help (@yandolfat was the brain behind all this ^^) !

There you go, the DCE chart has been updated as well ;)

BR

Mikaciu:
By the way @jCOTINEAU, we found out an unexpected behaviour during the startup of the app.

We have a java exception thrown by the liveness probe:
```ERROR web[AYwkaMbPoyiTKwPCAAAA][o.s.s.w.WebServiceEngine] Fail to process request http://172.18.39.55:9000/api/system/liveness```

It seems that the startupProbe is using `api/system/status` as endpoint, would it be possible to use `api/system/liveness` ?

The idea is coming from there : https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-startup-probes

> The trick is to set up a startup probe with the same command, HTTP or TCP check, with a failureThreshold * periodSeconds long enough to cover the worst case startup time.

rhass:
> We are investigating why it has not triggered the same failure on our EKS clusters, which is quite surprising.

I think a possible reason this is getting missed is because statefullsets have node affinity. Because of this, when and upgrade happens the node already has had the sysctl values applied and is able to come up without issue. When the node is replaced or an event happens which causes k8s to allocate the pod to a different node this behavior manifests itself. 

I was able force the behavior to happen by deleting the pod, and when it allocated to a different node in the cluster it failed to come up.

jCOTINEAU:
I have triggered a duplicated [PR](https://github.com/SonarSource/helm-chart-sonarqube/pull/401) in order to trigger the tests (we have an ongoing issue where an external contributor commit cannot run some part of the CI)

Could you also add the changelog entry to the chart.yml ?

Mikaciu:
> I have triggered a duplicated [PR](https://github.com/SonarSource/helm-chart-sonarqube/pull/401) in order to trigger the tests (we have an ongoing issue where an external contributor commit cannot run some part of the CI)
> 
> Could you also add the changelog entry to the chart.yml ?

Done ;)

Mikaciu:
> Hello @Mikaciu, The tests are failing because of ` * invalid changes annotation. Please use quotes on strings that include any of the following characters: {}:[],&*#?|-<>=!%@` coming from the `"Run the initSysctl init-container as root to prevent "permission denied" issues"` string.
> 
> Could you either remove the quotes or enclose it onto single quotes
> 
> After that, we should be all set, thanks again for your contribution !

Done ;)

jCOTINEAU (approved):
lgtm.

PR #402, opened 2023-12-15
Draft: BREAKING change for PDB typo

waynekhan:
1. Fixed the PodDistributionBudget typo.

2. I'll be happy to update the chart version; e.g., `11.0.0` pending your review per [SemVer](https://semver.org/).


jCOTINEAU:
Hello @waynekhan, thanks a lot for taking the time to participate in the community.

### Regarding point 1.
That is a very good catch and improvement, nonetheless, we try to deprecate things before breaking completely the behavior. Could you update your PR accordingly by:
- changing `values.schema.json` to express that a value is getting deprecated.
- change the logic so that it use the new `x.podDisruptionBudget` and rely on the old one if not found.

### Regarding point 2.

We explicitly decided to not follow Semver, the main reason is that we had a full draft between SQ versioning and helm chart versioning, which was really confusing for users and did not provide many benefits.

Also, the helm chart is released **only** after SQ is released. (we do have special cases in place for critical security patches)

We now consider the helm chart to be a sub-product of SonarQube, therefore we follow the same release cycle and versioning.

Hope this clarifies the situation, please let me know if you have other questions or would like to dig into the why of that decision.

charts/sonarqube-dce/CHANGELOG.md
```diff
@@ -2,7 +2,8 @@
 All changes to this chart will be documented in this file.
 
 ## [TBD]
-* Fix PDB typo
+
+* `sonarqube-dce`-only: Introduce `ApplicationNodes.podDistributionBudget` and `searchNodes.podDistributionBudget` in place of deprecated `ApplicationNodes.podDistributionBudget` and `searchNodes.podDistributionBudget`.
```
jCOTINEAU, line 6: this does not require to be prefixed by `sonarqube-dce-only` as it is already in the `sonarqube-dce` chart folder.

jCOTINEAU (commented):
overhall lgtm, the if else block is producing an extra blank line, which is not a big deal.

jCOTINEAU:
Hello @waynekhan huge thanks for your contribution! we completed the feature on top of your work [here](https://github.com/SonarSource/helm-chart-sonarqube/pull/411)

Closing as done

PR #410, opened 2024-01-11
SONAR-21408 run static test in PR

jCOTINEAU:


carminevassallo (commented):
Hey @jCOTINEAU is there a specific reason for running the static test in GH Actions rather than Cirrus?

jCOTINEAU:
@carminevassallo, comfort mostly as we already had a similar behavior for next scan with our GitHub_action, so it was very quick to implement.

This is a very small baby step in my opinion.

carminevassallo:
Given that we are trying to use one CI system, would that be better to spend a little bit more time but implement the step in Cirrus?

carminevassallo (changes requested):
Thanks for refactoring your previous change! I think we could add another small improvement that will be useful for the other tasks too. As of now, in our CI, we don't test against a specific version of k8s.

How about we add an ENV fixing the k8s version (e.g., `KUBE_VERSION`) that we use to execute the static tests and to create the kind cluster (in the `wait_for_kind` script we don't specify the k8s version)?

p.s.
I also suggest that we rename the task into "static_compatibility_test"

carminevassallo (approved):
Great job 💯

PR #411, opened 2024-01-16
Community PR to fix PDB typo on top of #402

jCOTINEAU:


charts/sonarqube-dce/templates/pod-disruption-budget.yaml
```diff
@@ -27,9 +32,15 @@ metadata:
     release: {{ .Release.Name }}
     heritage: {{ .Release.Service }}
 spec:
-{{- with .Values.ApplicationNodes.podDistributionBudget }}
-{{ toYaml . | indent 2 }}
-{{- end }}
+{{ if .Values.ApplicationNodes.podDistributionBudget }}
```
carminevassallo, line 35: Shouldn't be the other way round? If people define the `podDistruptionBudget`, this should have the precedence towards the deprecated value that is `podDistributionBudget`...

charts/sonarqube-dce/templates/pod-disruption-budget.yaml
```diff
@@ -9,13 +9,18 @@ metadata:
     release: {{ .Release.Name }}
     heritage: {{ .Release.Service }}
 spec:
-{{- with .Values.searchNodes.podDistributionBudget }}
-{{ toYaml . | indent 2 }}
-{{- end }}
+{{ if .Values.searchNodes.podDistributionBudget }}
+  {{- with .Values.searchNodes.podDistributionBudget -}}
```
carminevassallo, line 13: See my comment below...

carminevassallo (changes requested):

charts/sonarqube-dce/templates/pod-disruption-budget.yaml
jCOTINEAU: i dont think so.

In terms of use cases new user will have the proper one coming from the default values.

But people that use the deprecated one, will also get an updated default value, and that default value should not be used for them so the first thing to check is if the deprecated one is being defined.

wdyt ?

carminevassallo: Yes, you're right. If we change the order here, the users having the old value will get it overridden, which is not what we want to do when deprecating...

charts/sonarqube-dce/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ All changes to this chart will be documented in this file.
 * Improve the description of deprecated `ApplicationNodes.jvmOpts` and `ApplicationNodes.jvmCeOpts` values
 * Run the initSysctl init-container as root to prevent 'permission denied' issues
 * Add revisionHistoryLimit configuration for SonarQube application Deployment ReplicaSets
+* Introduce `ApplicationNodes.podDistributionBudget` and `searchNodes.podDistributionBudget` and deprecate `ApplicationNodes.podDistributionBudget` and `searchNodes.podDistributionBudget`.
```
carminevassallo, line 9: Small typo while deprecating a typo xD 

"Introduce `ApplicationNodes.podDistributionBudget` and `searchNodes.podDistributionBudget`" -> "Introduce `ApplicationNodes.podDisruptionBudget` and `searchNodes.podDisruptionBudget`"

charts/sonarqube-dce/Chart.yaml
```diff
@@ -35,6 +35,8 @@ annotations:
       description: "Run the initSysctl init-container as root to prevent 'permission denied' issues"
     - kind: changed
       description: "Add revisionHistoryLimit configuration for SonarQube application Deployment ReplicaSets"
+    - kind: deprecated
+      description: "Introduce `ApplicationNodes.podDistributionBudget` and `searchNodes.podDistributionBudget` and deprecate `ApplicationNodes.podDistributionBudget` and `searchNodes.podDistributionBudget`."
```
carminevassallo, line 39: see my comment above

carminevassallo (changes requested):

carminevassallo (approved):

carminevassallo:
LGTM, @jCOTINEAU! Don't forget to squash, I see quite a lot of commits

PR #428, opened 2024-02-22
DRAFT SONAR-21708 implement google cloud k8s sonarqube app

jCOTINEAU:


google-cloud-marketplace-k8s-app/README.md
```diff
@@ -0,0 +1,43 @@
...
+docker build -f google-cloud-marketplace-k8s-app/Dockerfile --tag $REGISTRY/$APP_NAME/deployer:$MINOR_VERSION .
+```
+
+With the deployer being built, one can deploy the app as is, with 
+
+```shell
+# make sure the namespace has been created already.
+export JWT_SECRET=$(echo -n "your_secret" | openssl dgst -sha256 -hmac "your_key" -binary | base64)
+
+mpdev install \
```
davividal, line 29: You can replace with this:

```shell
mpdev install \
  --deployer=$REGISTRY/$APP_NAME/deployer:$MINOR_VERSION \
  --parameters='{"name": "sonarqube-dce-gcapp-test", "namespace": "test-ns","ApplicationNodes.jwtSecret": "'"${JWT_SECRECT}"'", "postgresql.enabled": true, "jdbcOverwrite.enable": false }'
```

google-cloud-marketplace-k8s-app/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+### App verify
+
+On top of installing trought the deployer, we need to verify that our app comply with the requirements.
+
+```shell
+mpdev verify \
+  --deployer=$REGISTRY/$APP_NAME/deployer:$MINOR_VERSION \
+  --wait_timeout=1200 \
+  --parameters='{"name": "sonarqube-dce-gcapp-test", "namespace": "test-ns","ApplicationNodes.jwtSecret": "dZ0EB0KxnF++nr5+4vfTCaun/eWbv6gOoXodiAMqcFo=", "postgresql.enabled": true, "jdbcOverwrite.enable": false }'
```
davividal, line 42: Same as above.

davividal (changes requested):

google-cloud-marketplace-k8s-app/README.md
jCOTINEAU: true, but in the end those are for testing and we don't really need a dynamic token in this case,I would even suggest removing 
```
export JWT_SECRET=$(echo -n "your_secret" | openssl dgst -sha256 -hmac "your_key" -binary | base64)
```

wdyt ?

carminevassallo: I agree, we could remove, as it is not used..

google-cloud-marketplace-k8s-app/README.md
carminevassallo: We could remove the variable setup as it is not used.

charts/sonarqube-dce/templates/application.yml
```diff
@@ -0,0 +1,58 @@
+{{- if .Values.gcp_marketplace }}  # declared in google-cloud-marketplace-k8s-app value folder
```
carminevassallo, line 1: For now, we use the gcp_marketplace value only in the schema.yaml to activate this manifest. However, shall we document it anyways in the README and values.yaml of the chart? This might be useful to not forget about this value and have all the values clearly documented...

carminevassallo (changes requested):
Both install and verify work well. The PR is ready to be merged, just a few minor changes/questions.

carminevassallo (approved):
Looks great! Just a small reminder about not squashing all into one commit but keeping two, as you were addressing two tasks...

carminevassallo:
Hey @davividal, could you please review it too? There are requested changes that prevent this from being merged...

.cirrus/unit_helm_compatibility_test.sh
```diff
@@ -16,6 +16,6 @@ echo 'Running unit compatibility tests for Kubernetes version' $KUBE_VERSION
 for file in "$STATIC_TEST_FOLDER"/*; do
     TEST_CASE_NAME=$(basename "$file")
     echo 'Entering test for' $TEST_CASE_NAME
-    helm template --kube-version $KUBE_VERSION --dry-run --debug -f "$file" $TEST_CASE_NAME . | kubeconform --kubernetes-version $KUBE_VERSION --summary --strict
+    helm template --kube-version $KUBE_VERSION --dry-run --debug -f "$file" $TEST_CASE_NAME . | kubeconform --kubernetes-version $KUBE_VERSION --summary --strict -schema-location default -schema-location "../../tests/test-crds/app.k8s.io/application_v1beta1.json"
```
davividal, line 19: Is it `-schema-location` (single `-`) or `--schema-location` (double `-`)?

davividal (changes requested):
A small comment on the compatibility test. Other than that, LGTM.

.cirrus/unit_helm_compatibility_test.sh
jCOTINEAU: i think its single actually, but we have been using double. but looks like kubeconform handle both.

```
  -schema-location value
        override schemas location search path (can be specified multiple times)
```

davividal (approved):

PR #449, opened 2024-03-12
SONAR-21783 Set automountServiceAccountToken to false in pod's specif…

carminevassallo:


jCOTINEAU (changes requested):
looks all good to me, but i think we might be missing many places where we should put the automount to false.

All pod spec should have this, which include the job change-admin-password-hook.yml
 as well as the deployment.yml from the sonarqube chart

carminevassallo:
> looks all good to me, but i think we might be missing many places where we should put the automount to false.

I noticed that too. However, those issues were not flagged by the analyzer. So, I would like to open a dogfood post first to highlight that some important resources (such as deployments) are left out when are not set to be "templated" by default (which is, for example, the case of `deployment.yaml`).

jCOTINEAU (approved):
lgtm !

PR #450, opened 2024-03-12
SONAR-20656 change default memory chart settings to match default SQ …

jCOTINEAU:


charts/sonarqube/values.yaml
```diff
@@ -362,14 +362,15 @@ annotations: {}
...
 ## We usually don't make specific ressource recommandations, as they are heavily dependend on
 ## The usage of SonarQube and the surrounding infrastructure.
-## Adjust these values to your needs, but make sure that the memory limit is never under 4 GB
+## Adjust these values to your needs, but make sure that the memory limit is never under 6 GB
+## Those default are based on the default Web -Xmx1G -Xms128m and CE -Xmx2G -Xms128m and Search -Xmx2G -Xms2G settings of SQ sub processes
 resources:
   limits:
     cpu: 800m
-    memory: 4Gi
+    memory: 6144M
```
carminevassallo, line 370: Why increase it to 6Gi? Shouldn't we use 5Gi instead?

jCOTINEAU: The sum of all XMX is 5Gb and the recommendation is to have XMX at 80% of the total available memory.

This is why we always have a bit more kubernetes request/limit than the sum of all XMX.

charts/sonarqube/values.yaml
```diff
@@ -362,14 +362,15 @@ annotations: {}
...
   requests:
     cpu: 400m
-    memory: 2Gi
+    memory: 2048M
```
carminevassallo, line 373: The combined minimum heap size should be more that 2Gi...shall we set it to be 2560M (~2,5Gi)?

charts/sonarqube/values.yaml
```diff
@@ -362,14 +362,15 @@ annotations: {}
 
 ## We usually don't make specific ressource recommandations, as they are heavily dependend on
 ## The usage of SonarQube and the surrounding infrastructure.
-## Adjust these values to your needs, but make sure that the memory limit is never under 4 GB
+## Adjust these values to your needs, but make sure that the memory limit is never under 6 GB
```
carminevassallo, line 365: "is never under" -> "is never above"

charts/sonarqube-dce/values.yaml
```diff
@@ -105,14 +105,15 @@ searchNodes:
...
+  ## Those default are based on the default -Xmx2G -Xms2G settings of searchNodes sub processes.
   resources:
     limits:
       cpu: 800m
-      memory: 4096M
+      memory: 3072M
     requests:
       cpu: 400m
-      memory: 2Gi
+      memory: 3072M
```
carminevassallo, line 116: Why setting requests and limits to the same value here? I see that here and below, you try to set an extra Gi to the processes, what's the reason?

carminevassallo (commented):
Just added a few comments about the set limits/requests.

charts/sonarqube-dce/values.yaml
jCOTINEAU: So the reasoning is as follow.

For DCE chart we have full control so I suggest we go for request=limit which is much better for memory based application like java.

we are always adding an overhead of 1Gi compared to the sum of XMX and XMS cause those values are not a hard cap. They cap only some part of the JVM heap memory allocation so it is always recommended to have xmx/xms at 80% of the available memory (here materialized with request/limits).

Basically this is a strongly opinionated towards stability.

charts/sonarqube/values.yaml
jCOTINEAU: Here its a bit tricky as we need to comply with 3 different editions.

I would have prefered to setup 6G as request and limits, but this would probably break some community editions and this would be too big.

I am open to debate actually, i don't think we have a silver bullet for this.

charts/sonarqube/values.yaml
carminevassallo: thanks for clarifying!

charts/sonarqube-dce/values.yaml
carminevassallo: shall we add this recommendation in the readme for those cases? It can be added in parentheses next to the values (`xx.requests` and `xx.limits`) description.

charts/sonarqube-dce/README.md
```diff
@@ -130,6 +130,30 @@ Nonetheless, if you intend to run a production-grade SonarQube please follow the
 * Set `initSysctl.enabled` to **false**. This parameter would run **root** `sysctl` commands, while those sysctl-related values should be set by the Kubernetes administrator at the node level (see [here](#elasticsearch-prerequisites))
 * Set `initFs.enabled` to **false**. This parameter would run **root** `chown` commands. The parameter exists to fix non-posix, CSI, or deprecated drivers.
 
+#### Cpu and memory settings
+
+Monitoring cpu and memory is an important part of software reliability. The SonarQube helm chart comes with default values for cpu and memory requests and limits. Those memory values are matching the default Sonarqube JVM Xmx and Xms values.
```
joe-tingsanchali-sonarsource, line 135: A small request: can you match the stylization of SonarQube as `SonarQube` and not `Sonarqube`? There are a couple of places to correct.

charts/sonarqube-dce/README.md
```diff
@@ -130,6 +130,30 @@ Nonetheless, if you intend to run a production-grade SonarQube please follow the
...
+
+Xmx define the maximum size of the JVM heap, this is **not** the maximum memory the JVM can allocate.
+
+This is why most recommendation state that Xmx should be set to ~80% of the total amount of memory available on the machine.(or in kubernetes, the request and limits)
+
+Please find here the default Sonarqube Xmx parameter to setup the memory requests and limits accordingly.
+
+|Edition|Sum of Xmx|
+|---|---|
+|datacenter edition SearhNodes|2G|
```
joe-tingsanchali-sonarsource, line 145: Should this be `searchNodes` or `SearchNodes`? The values.yaml shows as `searchNodes`.

charts/sonarqube-dce/README.md
```diff
@@ -130,6 +130,30 @@ Nonetheless, if you intend to run a production-grade SonarQube please follow the
 * Set `initSysctl.enabled` to **false**. This parameter would run **root** `sysctl` commands, while those sysctl-related values should be set by the Kubernetes administrator at the node level (see [here](#elasticsearch-prerequisites))
 * Set `initFs.enabled` to **false**. This parameter would run **root** `chown` commands. The parameter exists to fix non-posix, CSI, or deprecated drivers.
 
+#### Cpu and memory settings
+
+Monitoring cpu and memory is an important part of software reliability. The SonarQube helm chart comes with default values for cpu and memory requests and limits. Those memory values are matching the default SonarQube JVM Xmx and Xms values.
+
+Xmx define the maximum size of the JVM heap, this is **not** the maximum memory the JVM can allocate.
```
carminevassallo, line 137: define -> defines

charts/sonarqube-dce/README.md
```diff
@@ -130,6 +130,30 @@ Nonetheless, if you intend to run a production-grade SonarQube please follow the
...
+
+This is why most recommendation state that Xmx should be set to ~80% of the total amount of memory available on the machine.(or in kubernetes, the request and limits)
```
carminevassallo, line 139: "This is why most recommendation state that Xmx should be set to ~80% of the total amount of memory available on the machine.(or in kubernetes, the request and limits)" -> "It is recommended to set `Xmx` to the ~80% of the total amount of memory available on the machine (in Kubernetes, this corresponds to requests and limits)."

charts/sonarqube-dce/README.md
```diff
@@ -130,6 +130,30 @@ Nonetheless, if you intend to run a production-grade SonarQube please follow the
...
+
+Please find here the default SonarQube Xmx parameter to setup the memory requests and limits accordingly.
```
carminevassallo, line 141: parameter -> parameters

charts/sonarqube-dce/README.md
```diff
@@ -130,6 +130,30 @@ Nonetheless, if you intend to run a production-grade SonarQube please follow the
...
+
+|Edition|Sum of Xmx|
+|---|---|
+|datacenter edition searchNodes|2G|
+|datacenter edition ApplicationNodes|3G|
+
+This is why to accomodate with the 80% rule we set the default to:
```
carminevassallo, line 148: "This is why to accomodate with the 80% rule we set the default to:" -> "To comply with the 80% rule mentioned above, we set the following default values:"

charts/sonarqube-dce/README.md
```diff
@@ -130,6 +130,30 @@ Nonetheless, if you intend to run a production-grade SonarQube please follow the
...
+
+- searchNodes.resources.memory.request/limit=3072M
+- ApplicationNodes.resources.memory.request/limit=4096M
+
+Please feel free to adjust those values to your needs but memory being a non-compresible resource, we advise to setup the memory requests and limits to the **same** value for production use case, making memory a guaranteed resource.
```
carminevassallo, line 153: Maybe change to ---> Please feel free to adjust those values to your needs. However, given that memory is a “non-compressible” resource, we advise you to set the memory requests and limits to the **same**, making memory a guaranteed resource. This is needed especially for production use cases.

charts/sonarqube-dce/README.md
```diff
@@ -130,6 +130,30 @@ Nonetheless, if you intend to run a production-grade SonarQube please follow the
...
+
+In order to change Xmx and Xms, prese refer to [this documentation](https://docs.sonarsource.com/sonarqube/latest/setup-and-upgrade/configure-and-operate-a-server/environment-variables/) and set the env_var or sonar.properties accordingly.
```
carminevassallo, line 155: To get some guidance when setting the Xmx and Xms values, please refer to [this documentation](https://docs.sonarsource.com/sonarqube/latest/setup-and-upgrade/configure-and-operate-a-server/environment-variables/) and set the environment variables or sonar.properties accordingly.

carminevassallo (changes requested):
I suggested some rephrasing of the README in `sonarqube-dce`. Could you please propagate your changes to the other README too?

carminevassallo (approved):

PR #451, opened 2024-03-13
Fix limits-related issues

carminevassallo:


charts/sonarqube/values.yaml
```diff
@@ -519,7 +521,10 @@ sonarqubeFolder: /opt/sonarqube
 tests:
   image: ""
   enabled: true
-  resources: {}
+  resources:
+    limits:
+      cpu: 800m
+      memory: 6144M
```
jCOTINEAU, line 527: For the test pods i think we can have request/limits much lower, as we only execute kind of a bash script that ping the APP-UI. This will not trigger the JVM creation process if I recall correctly, we overwrite the entrypoint.

charts/sonarqube-dce/README.md
```diff
@@ -211,8 +211,10 @@ The following table lists the configurable parameters of the SonarQube chart and
...
 | `searchNodes.resources.requests.cpu`                      | cpu request for Search Nodes                                                          | `400m`                                                                 |
+| `searchNodes.resources.requests.ephemeral-storage`        | storage request for Search Nodes                                                        | `1536M`                                                                |
 | `searchNodes.resources.limits.memory`                     | memory limit for Search Nodes. should not be under 4G                                 | `4096M`                                                                |
 | `searchNodes.resources.limits.cpu`                        | cpu limit for Search Nodes                                                            | `800m`                                                                 |
+| `searchNodes.resources.limits.ephemeral-storage`          | storage limit for Search Nodes                                                        | `512000M`                                                                |
```
jCOTINEAU, line 217: 512G for the ephemeral limit ? is there a specific reason for that ?

jCOTINEAU (changes requested):

charts/sonarqube/values.yaml
carminevassallo: yes, we just do a wget or something similar.  This is a fair point because if we don't specify the requests, the limits we set will become requests too and this is not what I wanted...lowering the limits now :)

charts/sonarqube-dce/README.md
jCOTINEAU: We will have configmap, secret, volumes counting towards this.

For us the biggest would be plugins ? that are stored under some volumes and can weight few gigs of JAR.

Dont have a strong opinion, i am fine with keeping that way.

carminevassallo: Ok, I'll keep like this, without setting any strict limit..

jCOTINEAU (approved):
lgtm !

PR #455, opened 2024-03-14
SONAR-21775 Rename tasks in Helm chart CI

davividal:


.cirrus/tasks.yml
```diff
@@ -233,24 +246,24 @@ trigger_release_task:
       - ./.cirrus/slack-notification.sh
 
 build_gcp_prod_app_task:
-  <<: *BUILD_GCP_STAGING_APP_TASK
+  <<: *BUILD_GCP_APP_TEMPLATE
   only_if: $TRIGGER == "PUSH_PROD_DEPLOYER"
   env:
-    GCLOUD_PRODUCT_NAME: sonarqube-dce-staging
+    GCLOUD_PRODUCT_NAME: official-sonarqube-data-center-edition
```
carminevassallo, line 252: We should still build in our staging account. When we want to release a new image to the prod project, we need to build and verify in the staging account. So, we can reuse the "build_gcp_staging_app" and "verify_gcp_staging_app" tasks when we want to trigger "PUSH_PROD_DEPLOYER" and remove both  "build_gcp_prod_app" and "verify_gcp_prod_app" tasks as they are redundant.

.cirrus/tasks.yml
```diff
@@ -233,24 +246,24 @@ trigger_release_task:
...
-  <<: *BUILD_GCP_STAGING_APP_TASK
+  <<: *BUILD_GCP_APP_TEMPLATE
   only_if: $TRIGGER == "PUSH_PROD_DEPLOYER"
   env:
     GCLOUD_REGISTRY: gcr.io/sonarsource-public
     GCLOUD_PRODUCT_NAME: official-sonarqube-data-center-edition
   depends_on:
-    - verify_gcp_prod_app
\ No newline at end of file
+    - verify_gcp_prod_app
```
carminevassallo, line 269: If we make the changes I mentioned above, we should make this task dependent on verify_gcp_staging_app

carminevassallo (changes requested):

carminevassallo (approved):

PR #459, opened 2024-04-04
SONAR-21969 remove outdated readme doc about persistency

jCOTINEAU:
This part of the doc is actually a leftover from the fork between standard and dce. 

The correct section is already described under SearchNodes.persistency


davividal (changes requested):
Does it also make sense to replicate this to the `sonarqube` chart?

jCOTINEAU:
> Does it also make sense to replicate this to the `sonarqube` chart?

It exist only on the dce, on the standard one this block is the correct one, as persistency is a top value field

davividal (approved):
LGTM!

PR #461, opened 2024-04-08
SONAR-20106 deprecate nginx field in favor of ingress-nginx subchart …

jCOTINEAU:


charts/sonarqube-dce/values.schema.json
```diff
@@ -1,6 +1,11 @@
 {
     "$schema": "https://json-schema.org/draft/2020-12/schema",
     "properties": {
+        "nginx.enabled": {
```
carminevassallo, line 4: I think we should deprecate nested values as we did for example for "readinessProbe.sonarWebContext"

charts/sonarqube/values.schema.json
```diff
@@ -4,6 +4,11 @@
         "replicaCount"
     ],
     "properties": {
+        "nginx.enabled": {
```
carminevassallo, line 7: see my comment above...

carminevassallo (changes requested):
Validate it locally too. It works great!

A minor comment on the values schema and an open question. I see that, in the past, when deprecating a value instead of removing it from the `values.yaml` we just commented it, explaining that this was deprecated (see what we did for `account. sonarWebContext`). Shall we adopt the same pattern?

carminevassallo (approved):
LGTM! 🎉

PR #464, opened 2024-04-09
SONAR-20266 fix duplicated env var caused by templating

jCOTINEAU:


charts/sonarqube-dce/templates/_helpers.tpl
```diff
@@ -294,4 +294,23 @@ Set sonarqube.webcontext, ensuring it starts and ends with a slash, in order to
...
+
+
+{{/*
+Set combined_env, ensuring we dont have any duplicates with our features and some of the user provided env vars
+*/}}
+{{- define "sonarqube.combined_app_env" -}}
+{{- $filteredEnv := list -}}
+{{- range $index,$val := .Values.ApplicationNodes.env -}}
+# we remove every env var that are already defined by the functions in this tpl file or by other features
+  {{- if has $val.name (list "SONAR_WEB_CONTEXT" "SONAR_WEB_JAVAOPTS" "SONAR_CE_JAVAOPTS" "SONAR_JDBC_PASSWORD" "SONAR_JDBC_PASSWORD" "SONAR_WEB_SYSTEMPASSCODE") -}}
```
jCOTINEAU, line 307: here we could invert the if order, having a if not has do

jCOTINEAU:
With the current setup, the code produce such vars:
```
            - name: SONAR_WEB_JAVAOPTS
              value: '"-Djavax.net.ssl.trustStore=/opt/sonarqube/certs/cacerts my-value"'
            - name: SONAR_CE_JAVAOPTS
              value: '"-Djavax.net.ssl.trustStore=/opt/sonarqube/certs/cacerts"'
```

is the double quoting an issue in yaml/java ?

jCOTINEAU:
> With the current setup, the code produce such vars:
> 
> ```
>             - name: SONAR_WEB_JAVAOPTS
>               value: '"-Djavax.net.ssl.trustStore=/opt/sonarqube/certs/cacerts my-value"'
>             - name: SONAR_CE_JAVAOPTS
>               value: '"-Djavax.net.ssl.trustStore=/opt/sonarqube/certs/cacerts"'
> ```
> 
> is the double quoting an issue in yaml/java ?
> 
> This was not observed before, it might be caused by the toJson / fromJson / toYaml situation.

Has been fixed in the second commit, its adding a bit of overhead but still looks good.

carminevassallo (approved):
L(super)GTM!

Really nice way to handle such a complex workflow! 👏

jonesbusy:
This cause : https://community.sonarsource.com/t/upgrade-from-10-4-1-to-10-5-0-broke-ldap-authentication/113722/9

What about user passing env vars from secret ?

```
  - name: LDAP_****2****BINDPASSWORD
    valueFrom:
      secretKeyRef:
        name: "sonarqube-ldap"
        key: LDAP_BINDPASSWORD
```

You assume there is always a 'value' which is not always the case

PR #491, opened 2024-06-12
SONAR-22353 Define the HorizontalPodAutoscaler for the app deployment

carminevassallo:


jCOTINEAU (changes requested):
Super cool improvement!

I think we should also change the deployment behavior as described [here](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#migrating-deployments-and-statefulsets-to-horizontal-autoscaling) to make sure that on deploy we do not overwrite the hpa count.

charts/sonarqube-dce/values.yaml
```diff
@@ -344,6 +344,32 @@ ApplicationNodes:
...
+        policies:
+          - type: Pods
+            value: 1
+            periodSeconds: 0
+      scaleUp:
+        stabilizationWindowSeconds: 60
+        policies:
+          - type: Pods
+            value: 1
+            periodSeconds: 0
```
jCOTINEAU, line 371: also when trying to install it by only changing `enabled:true` I get this error
`Error: UPGRADE FAILED: failed to create resource: HorizontalPodAutoscaler.autoscaling "toto-sonarqube-dce-app-hpa" is invalid: [spec.behavior.scaleUp.policies[0].periodSeconds: Invalid value: 0: must be greater than zero, spec.behavior.scaleDown.policies[0].periodSeconds: Invalid value: 0: must be greater than zero] `

jCOTINEAU (approved):
lgtm! 

Small question, should we add the link to the official doc afterward ?

carminevassallo:
> Small question, should we add the link to the official doc afterward ?

Yup, we discussed this, we are gonna update the doc when ready.

PR #492, opened 2024-06-18
fix(BUILD-5362): add eks access

hedinasr:
We want to drop the usage of Docker in Docker. See https://discuss.sonarsource.com/t/build-5170-support-developers-to-migrate-from-usage-of-docker-in-docker/19010.


.cirrus/tasks.yml
```diff
@@ -140,25 +140,16 @@ chart_testing_task:
...
-        privileged: true
-        cpu: 4
-        memory: 16Gb
-        image: public.ecr.aws/docker/library/docker:dind
-        port: 2375
-        env:
-          DOCKER_DRIVER: overlay2
-          DOCKER_TLS_CERTDIR: ""
   <<: *CLONE_SCRIPT_TEMPLATE
   start_kind_script:
```
carminevassallo, line 144: We are now connecting to a real test cluster instead of an emulated one :) Shall we rephrase the task's name with something like "authenticate_to_test_cluster"? :)

carminevassallo (changes requested):
Hey @hedinasr,

This is a super cool improvement to our CI process!!!

We have been struggling a lot with using an emulated cluster and I'm sure connecting to a real one will prevent all the flakiness we saw from happening.

Suggesting just a small change of the task's name and then it should be ready to be merged :)

.cirrus/tasks.yml
hedinasr: We can let the name of the task as it is because we didn't change what the task is doing.

carminevassallo: sorry, not the overall task's name, just the step's name to something like "authenticate_to_test_cluster_script". WDYT?

carminevassallo: also, we should delete the `wait_forkind.sh`... this script won't be used anymore

hedinasr: oh yes totally, I will make those changes. Thank you.

PR #494, opened 2024-06-20
SONAR-22334 makes readiness probe and liveness probe configurable

carminevassallo:


jCOTINEAU (approved):
LGTM, validated with legacy parameters

PR #497, opened 2024-06-21
SONAR-22442 Prevents sonarWebContext null exception

davividal:


davividal:
This is actually not a bug, as pointed out in the related ticket.

PR #500, opened 2024-06-24
SONAR-18996 Deprecate jdbcOverwrite.enable in favor of jdbcOverwrite.…

jCOTINEAU:


carminevassallo (changes requested):
Great PR! It's awesome that we finally deprecate this typo :)

I validated it functionally. It works as expected :tada:

Just a few minor points before we merge.
1. As it happens with other deprecated values, we can't just remove them from the parameters in the README.md (see `readinessProbe.sonarWebContext`). Can you please add the depreacate parameter mentioning that it's "(DEPRECATED)"? (same comment for both charts) Furthermore, instead of replacing the value in the `values.yaml`, we should comment it out and mentioned that it's deprecated in favor of the others.
2. In the DCE chart, the same changelog entry is present both in 10.7 and 10.6. We should fix this.

davividal (approved):
LGTM!

carminevassallo (approved):

PR #509, opened 2024-07-02
SONAR-22465 Skips helm packaging for PRs

davividal:


.cirrus/package.sh
```diff
@@ -1,20 +1,28 @@
...
+
+: "${CIRRUS_BRANCH:?}"
+: "${BUILD_NUMBER:?}"
+: "${CIRRUS_BASE_BRANCH:=}"
 
 VERSION_SEPERATOR="-"
 [[ ${CIRRUS_RELEASE:-} != "" ]] && VERSION_SEPERATOR="+"
 
-PREVIOUS_RELEASE=$(gh api "/repos/{owner}/{repo}/releases" --jq "[.[] | select(.target_commitish==\"${CIRRUS_BRANCH}\")][1].tag_name")
+[[ -n "${CIRRUS_BASE_BRANCH}" ]] && TARGET_BRANCH="${CIRRUS_BASE_BRANCH}" || TARGET_BRANCH="${CIRRUS_BRANCH}"
```
jCOTINEAU, line 12: Not sure to understand what this checks does and why we need it now ?

davividal: If this is a pull request, it has a base branch. In that case, I want to check if there are changes in the helm chart against the base branch, not the PR branch.

Otherwise, if this is another branch (release, LTA, etc.), I need to use that branch as a comparison.

We need the base branch to check the latest release for that branch.

It is not that we need it **now**, but we became aware of this bug in the last release. Previously, all comparisons were made against `master`, which is not the behavior we want 100% of the time.

jCOTINEAU (approved):
lgtm, cool imrpvoment!

PR #517, opened 2024-07-15
SONAR-22530 implement simple fixtures

jCOTINEAU:


carminevassallo (approved):
LGTM!

(I agree with you about not adding entries to the changelog about this change)

PR #523, opened 2024-07-22
SQDEVOPS-527 Set up slack notification of failures

carminevassallo:


jCOTINEAU (commented):
Small question, should it be on-failure only like in cirrus, or is the action smart enough to detect failures ?

carminevassallo:
Yup, in [the parent action](https://github.com/SonarSource/gh-action_build-notify/blob/master/.github/workflows/main.yaml), we see which are the statuses where the notification is not sent. So, yes we notify failures only...

jCOTINEAU (approved):
lgtm.

PR #525, opened 2024-07-25
SONAR-19441 enable read only root filesystem

jCOTINEAU:


carminevassallo (approved):
Tested locally too. It works as expected.

PR #526, opened 2024-08-02
SONAR-22588 Implements OpenShift validator

davividal:


carminevassallo (commented):
Hey @davividal,

when trying to verify the charts locally I have a different behaviour

- I don't get the chart-testing related violation that we see on Cirrus
- the `sonarqube-dce` verification does not succeed because the jwtSecret is missing.

Did you have similar issues? How is that possible that we don't get the jwtSecret issue when verifying the chart on Cirrus?

carminevassallo:
@davividal the build agent does not contain `kubectl`, we should fix this.

Anyways, apart from this issue, we have identified a bug in the verifier: we can't specify custom values to the chart during the installation (i.e., via `-S`). This makes the verify stage useless because we can't use the image under development (`10.7`) where we add Openshift annotations, and we can't add the jwt secret required for the dce to run. 

I'm proposing to close this PR/ticket, raise an issue on the [chart-verifier](https://github.com/redhat-certification/chart-verifier/) repo, and reopen this task once we get answer. WDYT?

davividal:
This PR is blocked by https://github.com/redhat-certification/chart-verifier/issues/465.

charts/sonarqube/ci/openshift-verifier-values.yaml
```diff
@@ -0,0 +1,8 @@
+jwtSecret: "dZ0EB0KxnF++nr5+4vfTCaun/eWbv6gOoXodiAMqcFo="
+
+image:
+  image:
```
komish, line 4: I think this key is extraneous.

komish:
> This PR is blocked by [redhat-certification/chart-verifier#465](https://github.com/redhat-certification/chart-verifier/issues/465).

This should be resolved.

jCOTINEAU:
Hello everyone, closing the PR for now, will reopen when we start the following sprint on OpenShift compatibility, thanks all for participating

.cirrus/Dockerfile
```diff
@@ -95,6 +97,19 @@ RUN set -eux; \
...
 
+RUN set -eux; \
+    curl -LO https://github.com/redhat-certification/chart-verifier/releases/download/${CHART_VERIFIER_VERSION}/chart-verifier-${CHART_VERIFIER_VERSION}.tgz ; \
+    echo "${CHART_VERIFIER_CHECKSUM}  chart-verifier-${CHART_VERIFIER_VERSION}.tgz" | sha256sum -c ; \
+    tar -xf chart-verifier-${CHART_VERIFIER_VERSION}.tgz ; \
+    chmod +x ./chart-verifier ; \
+    mv ./chart-verifier /usr/bin/chart-verifier ; \
+    rm chart-verifier-${CHART_VERIFIER_VERSION}.tgz
+
+RUN curl -LO https://github.com/derailed/k9s/releases/download/v0.32.5/k9s_Linux_amd64.tar.gz && \
```
jCOTINEAU, line 108: Is it still needed after debug/development phase ?

davividal: Ops! Good catch! I forgot this.

charts/sonarqube/ci/openshift-verifier-values.yaml
```diff
@@ -0,0 +1,14 @@
+OpenShift:
+  enabled: false
```
carminevassallo, line 2: Shall we set this to true?

davividal: If we do, the installation will fail because we are not running the verifier against an Openshift cluster. I'm using the CI cluster.

I think that using an Openshift cluster for this verification is overkill. WDYT?

carminevassallo: sure but shall we then remove the values from the openshift-verify ones? Openshift is disabled by default...

jCOTINEAU: > If we do, the installation will fail because we are not running the verifier against an Openshift cluster. I'm using the CI cluster.
> 
> I think that using an Openshift cluster for this verification is overkill. WDYT?

I think that conceptually it makes sense to run against an openshift cluster considering we are running the openshift specific verifier. I am not sure what tests/verifications are being done so its really an I speak I gut feeling

PR #533, opened 2024-08-22
SONAR-18992 fix network policy typo

jCOTINEAU:


charts/sonarqube/templates/networkpolicy.yaml
```diff
@@ -99,4 +99,7 @@ spec:
 {{- with .Values.networkPolicy.additionalNetworkPolicys -}}
 {{ toYaml . | nindent 2 }}
 {{- end }}
+{{- with .Values.networkPolicy.additionalNetworkPolicies -}}
```
carminevassallo, line 102: In this case, if we use both ways to add network policies, we end up with having the unsupported network policy and the support ones both added.

Shouldn't we "disable" `additionalNetworkPolicys` when `additionalNetworkPolicies` is used?

carminevassallo, line 102: In this case, if we use both ways to add network policies, we end up with having the unsupported network policy and the support ones both added.

Shouldn't we "disable" `additionalNetworkPolicys` when `additionalNetworkPolicies` is used?

carminevassallo (changes requested):

charts/sonarqube/templates/networkpolicy.yaml
jCOTINEAU: i thought about it, my reasoning so far is that the template is clean, no if else. In the end we are talking about adding stuff, so they can be added simultaneously it will not interfer with each other as its just a typo, and from the user need, they will very likely use one or the other.

But yeah maybe forcing the disable would be better I don't have a super strong opinion here

carminevassallo: I think we should be coherent with what we do elsewhere. If the new feature is used, the deprecated one is disabled...WDYT?

charts/sonarqube-dce/templates/networkpolicy.yaml
```diff
@@ -181,7 +181,13 @@ metadata:
     release: {{ .Release.Name }}
     heritage: {{ .Release.Service }}
 spec:
+{{- if.Values.networkPolicy.additionalNetworkPolicys -}}
```
carminevassallo, line 184: We should revert the order here. If people set the new value `.Values.networkPolicy.additionalNetworkPolicies` this should take precedence over the deprecated one, right?

carminevassallo (changes requested):

charts/sonarqube-dce/templates/networkpolicy.yaml
jCOTINEAU: Hm i usually prefer the other way around for compatibility

carminevassallo: Ok this seems to be the standard in our charts. Thanks for clarifying :)

carminevassallo (approved):

PR #536, opened 2024-08-23
SONAR-19664 support HTTPS traffic between app and search nodes

carminevassallo:


jCOTINEAU (commented):
after doing the review and looking at the feature I think we might have other bugs regarding how the useInternalSecret works, should we take the time to discuss it furthermore?

charts/sonarqube-dce/templates/_helpers.tpl
```diff
@@ -223,16 +223,27 @@ Create the name of the service account to use
 {{- end -}}
 
 {{/*
-Set search.useInternalKeystoreSecret
+Set searchAuthentication.useInternalKeystoreSecret when earchNodes.searchAuthentication.keyStorePasswordSecret is not provided
```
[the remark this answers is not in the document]
carminevassallo: The `useInternalKeystoreSecret` should be set to true when the `keyStorePassword` is provided and not the `keyStorePasswordSecret`

charts/sonarqube-dce/README.md
```diff
@@ -215,64 +215,83 @@ kubectl apply -f https://github.com/kubernetes-sigs/metrics-server/releases/late
 
 ```
 
+## Secure the communication within the cluster
+
+In order to secure the communication between Application and Search nodes, you need to set both `nodeEncryption.enabled` and `searchNodes.searchAuthentication.enabled` to `true`.
+
+In a secured cluster, Elasticsearch nodes use certificates to identify themselves when communicating with other nodes. You need to generate a Certificate Authority (CA) together with a certificate and private key for the nodes in your cluster. Furthemore, you need to specify the Search nodes' hostnames that will be added as DNS names in the Subject Alternative Name (SAN).
+
+As an example, let's assume that your cluster has three search nodes with the release's name set to "sq" and the chart's name set to "sonarqube-dce". You will need to add the following DNS names in the SAN.
```
jCOTINEAU, line 224: we might want to add that there is the namespace involved also ? and clearly show what field does which ?

because I think the pattern is release-sonarqube-dce-search-number.release-sonarqube-dce-search.namespace.svc.cluster.local.

which maps to pod.service.namespace.svc.cluster.local ?

Also we should specify that the last one is mandatory and is just the service, which is releaseName-sonarqube-dce-search

jCOTINEAU: and after doing it i think we should even provide a detailed tutorial on how to generate the cert

jCOTINEAU (approved):
lgtm, super cool security improvment

PR #539, opened 2024-08-26
SONAR-21136 Ensure cURL inputs are URL-encoded

davividal:


jCOTINEAU (approved):
lgtm, working as expected !

PR #557, opened 2024-09-25
SONAR-23150 Group all Openshift-specific resources under the same parameter

carminevassallo:


jCOTINEAU (commented):
Hey Carmine, nice improvment ! to help testing i would suggest updating the `openshift-values.yaml` to include the route settings.

To do so, you can generate the files from master and rerun from your modifications, and we should have no differences.

carminevassallo:
> to help testing i would suggest updating the `openshift-values.yaml` to include the route settings.

This makes a lot of sense, thanks for the suggestion!

jCOTINEAU (approved):
all good to me !

PR #558, opened 2024-09-25
SONAR-20820 Remove socketTimeout from jdbcOverwrite.jdbcUrl's default value

carminevassallo:


jCOTINEAU (approved):
all good to me for dogfood, side question, do we know which is the default now ?

PR #559, opened 2024-09-25
SONAR-22219 add extra volume for dce ApplicationNodes

jCOTINEAU:


charts/sonarqube/values.yaml
```diff
@@ -469,8 +471,10 @@ persistence:
   uid: 1000
   guid: 0
 
+  ## DEPRECATED please use root level extraVolumes value
```
carminevassallo, line 474: We should add `(DEPRECATED)` in the readme's descriptions of those parameters together with a small suggestion on what to use instead.

carminevassallo (approved):
The functionalities work as expected. A small comment remains. After fixing that, the PR can be merged!

carminevassallo (approved):
Great job!

PR #562, opened 2024-09-30
SONAR-19541 Enable users to install an oracle jdbc driver

carminevassallo:


charts/sonarqube/templates/_pod.tpl
```diff
@@ -60,8 +60,20 @@ spec:
...
+        - |
+          #!/bin/sh
+          cp -f "${JAVA_HOME}/lib/security/cacerts" /tmp/certs/cacerts
+          if [ "$(ls /tmp/secrets/ca-certs)" ]; then
+            for f in /tmp/secrets/ca-certs/*; do
+              keytool -importcert -file "${f}" -alias "$(basename "${f}")" -keystore /tmp/certs/cacerts -storepass changeit -trustcacerts -noprompt
+            done
+          fi
+          keytool -importkeystore -srckeystore /tmp/certs/cacerts -destkeystore /tmp/certs/cacerts.p12 -deststoretype PKCS12 -srcstorepass changeit -deststorepass changeit
+          openssl pkcs12 -info -in /tmp/certs/cacerts.p12 -out /tmp/certs/cacerts.pem -nodes -passin pass:changeit
```
jCOTINEAU, line 75: just to confirm, here the goal is to duplicate the created cert bundle but export it as pem format so that it can be used by curl?

Side question, Curl accept no other format ?
Side-side question: for java, we need a full structure, but I think for curl, the certificate will be an extra one on top of the system one, meaning we probably don't need the whole copy import logic. wdyt ? That might simplify the code

charts/sonarqube/templates/_pod.tpl
```diff
@@ -411,6 +448,14 @@ spec:
         items:
           - key: install_plugins.sh
             path: install_plugins.sh
+    {{- end }}
+        {{- if .Values.jdbcOverwrite.oracleJdbcDriverURL }}
```
jCOTINEAU, line 452: few extra space on the left side of the if ?

charts/sonarqube/templates/init-fs.yaml
```diff
@@ -12,6 +12,9 @@ data:
     {{- if .Values.plugins.install }}
     chown -R {{ .Values.persistence.uid }}:{{ .Values.persistence.guid }} {{ .Values.sonarqubeFolder }}/extensions/plugins
     {{- end }}
+    {{- if .Values.jdbcOverwrite.oracleJdbcDriverURL }}
+    chown -R {{ .Values.persistence.uid }}:{{ .Values.persistence.guid }} {{ .Values.sonarqubeFolder }}/extensions/jdbc-driver/oracle
```
jCOTINEAU, line 16: not tested yet, but does this work knowing that this path might not have been created yet? ha maybe yes because on the volume path definition it will automatically create the subpath ? will test

charts/sonarqube/templates/install-oracle-jdbc-driver.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ template "sonarqube.fullname" . }}-install-oracle-jdbc-driver
+  labels: {{- include "sonarqube.labels" . | nindent 4 }}
+data:
+  install_oracle_jdbc_driver.sh: |-
+    {{- if .Values.jdbcOverwrite.oracleJdbcDriverURL }}
```
jCOTINEAU, line 8: i think we should put the If on top of the file, not creating the configmap if its not needed, we have many places where we create resources that will be unused.

And because we have if on all mount part it should not be a risk

charts/sonarqube/templates/_pod.tpl
carminevassallo: Question 1/2: Not sure about all formats it supports. For sure, it does not support the keystore made for Java apps (JKS)

Question 3: What would be your suggestion here? Just add the file and load in the chart when needed?

charts/sonarqube/templates/_pod.tpl
carminevassallo: Addressed

charts/sonarqube/templates/install-oracle-jdbc-driver.yaml
carminevassallo: Addressed

charts/sonarqube/templates/_pod.tpl
carminevassallo: After discussing it, we decided not to convert the keystore but just add the certificate we need when downloading the plugin.

davividal (approved):
LGTM!

charts/sonarqube/templates/init-fs.yaml
davividal: I tested it and it doesn't seem to be a problem. :)

9numbernine9:
Hello! Just a passer-by that happens to be having problems with installing the Oracle driver when deploying `10.7.0`!

As part of this PR, would it be possible to add functionality to provide credentials when downloading the Oracle driver? (This could be similar to the `netrc` functionality that currently exists for plugins.) To provide some context, we have Sonarqube deployed in an airgapped environment so we can't download the driver from public unauthenticated mirrors, and the mirror we _do_ have within the airgap environment _requires_ authentication.

carminevassallo:
Hello @9numbernine9 👋 

> As part of this PR, would it be possible to add functionality to provide credentials when downloading the Oracle driver? 

Thanks for the suggestion. We discussed and gonna add this feature to the current PR 💯 

On a side note. The 10.7 charts are out already. For the DCE edition, we added the possibility to configure extraVolumes. This can be used as a workaround to pre-download the driver in an external volume in the cluster and mount it in the right folder. Feel free to leverage our [community](https://community.sonarsource.com/) to post issues that you might encounter.

The current PR's changes will get shipped with 10.8.

PR #564, opened 2024-10-01
REL-303 release helm charts 10.7.0

jCOTINEAU:


charts/sonarqube-dce/Chart.yaml
```diff
@@ -79,6 +79,8 @@ annotations:
       description: "Support ApplicationNodes.extraVolumes and ApplicationNodes.extraVolumeMounts"
     - kind: added
       description: "Ensure kubernetes.io/version label is smaller than 63 chars"
+    - kind: changed
```
carminevassallo, line 82: Based on our release process, we should add this change on top of the list (after the chart bump). However, I'm not sure I agree with doing it, wdyt?

carminevassallo (approved):
LGTM! Just a minor point to discuss...

charts/sonarqube-dce/Chart.yaml
jCOTINEAU: Agree, resolved

PR #565, opened 2024-10-01
REL-302 Release helm chart for SonarQube 9.9.7

jCOTINEAU:


carminevassallo (changes requested):
We need to update the `artifactub.io/changes` the in dce chart.

carminevassallo (approved):
LGTM!

PR #576, opened 2024-11-04
SONAR-23305 Test the helm charts with a local cluster

carminevassallo:


jCOTINEAU (approved):
lgtm, we could save a bit by having a VM already set, but 1 min is acceptable as a baby step nice and smooth shift

PR #577, opened 2024-11-08
SONAR-23558 Introduce the SonarQube Community Build support

carminevassallo:


jCOTINEAU (dismissed):
Super cool improvments, while i finalize the review and the local test, could you add some fixtures test cases to validate the behavior of the new image.tag helper template ?

jCOTINEAU (commented):
Should we specify somehow that the `community.enabled` and `edition` fields will become mutually exclusive? at first glance from user's point of view it's hard to know what happens if I put `community.enabled=true` and `edition=developer`

Nonetheless it might not be a real question considering that in the next release edition will be empty and community set to false.

Also right now edition is required but this will not be true as `edition=null` and `community.enabled=true` is a valid combination

Should we also start putting the fail-fast mechanism in case a wrong set of combinations is given instead of deploying something fishy?

davividal (approved):
LGTM!

PR #574, opened 2024-10-29
Add K8s Recommended Labels to Pods

wyattfry:
Please ensure your pull request adheres to the following guidelines:
- [x] explain your motives to contribute this change: what problem you are trying to fix, what improvement you are trying to make
- [x] Document your Changes in the `CHANGELOG.md` file of the respected chart as well as the `Chart.yaml`

This PR adds the Kubernetes-recommended labels "app.kubernetes.io/version" and "app.kubernetes.io/name" to the Pod template.

https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels

PR #585, opened 2024-11-14
SONAR-23624 Add K8s Recommended Labels to Pods

carminevassallo:
This PR is based on #574

### DIFF
--- a/charts/sonarqube/templates/_helpers.tpl
+++ b/charts/sonarqube/templates/_helpers.tpl
@@ -34,7 +34,7 @@ Selector labels
 {{- define "sonarqube.selectorLabels" -}}
 app: {{ include "sonarqube.name" . }}
 release: {{ .Release.Name }}
-app.kubernetes.io/name: {{ include "sonarqube.name" . }}-{{ include "sonarqube.fullname" . }}
+app.kubernetes.io/name: {{ .Release.Name }}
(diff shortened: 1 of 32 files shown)


charts/sonarqube/templates/_helpers.tpl
```diff
@@ -34,14 +34,16 @@ Selector labels
 {{- define "sonarqube.selectorLabels" -}}
 app: {{ include "sonarqube.name" . }}
 release: {{ .Release.Name }}
+app.kubernetes.io/name: {{ .Release.Name }}
```
davividal, line 37: Doesn't make sense to replicate this to `sonarqube-dce` as well?

davividal (approved):
If this is exclusive to `sonarqube`, LGTM!

charts/sonarqube/templates/_helpers.tpl
wyattfry: Hello, original author here, I'm open to adding the same changes to DCE. I didn't because i wasn't aware of the chart, my org only uses the non-DCE version.

carminevassallo: @davividal, we have already the label used in DCE :) So no need to ask @wyattfry for extra help on this ;)

wyattfry:
woohooo thanks all!

PR #587, opened 2024-11-15
SONAR-23627 improve initsysctl fail logic trigger 572

jCOTINEAU:


charts/sonarqube-dce/values.yaml
```diff
@@ -570,7 +570,7 @@ networkPolicy:
   # additionalNetworkPolicies:
 
 initContainers:
-  # image: busybox:1.36
+  # image: ubuntu:24.04
```
carminevassallo, line 573: We should set this value not to be `ubuntu` (even if commented). This might be a wrong suggestion, as the default value is the usual `sonarqube`. I would just leave `# image: `.

jCOTINEAU: hmm i agree, the goal there was to show you can change the image, but it should be explicit that it is not required.
I'll add the ubuntu above the image in a dedicated comment.

carminevassallo (approved):
Great improvement!

We notice that the conversion to `int` introduces a small inaccuracy in the case of very big numbers. However, we are talking about cases that are exceptional and that we are considering to "forbid" by not letting users modify the default parameters required by Elasticsearch.

PR #590, opened 2024-11-21
SONAR-23713 deprecate parameters in sonarqube chart

jCOTINEAU:


charts/sonarqube/values.yaml
```diff
@@ -604,8 +604,7 @@ postgresql:
 # podLabels:
 #   key: value
 podLabels: {}
-# For compatibility with 8.0 replace by "/opt/sq"
-# For compatibility with 8.2, leave the default. They changed it back to /opt/sonarqube
+# (DEPRECATED) this field will be removed, as it needs to match the SonarQube image folder structure, considering we have one chart version per docker image version this field is not needed anymore.
```
davividal, line 607: There is no "(DEPRECATED)" for the `terminationGracePeriodSeconds` in the values.yaml. Do you think it is necessary?

What about adding something to the `NOTES.txt`?

davividal (approved):
Despite my comment, I think this is good to go.

LGTM!

charts/sonarqube/values.yaml
jCOTINEAU: ha right good catch, i'll add this.

The notes seems a bit overkill as this is for things that user needs to act on. here nothing is required for terminationGracePeriod.

PR #593, opened 2024-11-22
SONAR-23714 Deprecate account in favor of setAdminPassword

carminevassallo:


charts/sonarqube-dce/values.schema.json
```diff
@@ -48,15 +48,47 @@
             }
         },
         "account": {
+            "type": "object",
+            "deprecated": true,
+            "$comment": "(DEPRECATED) Please use `setAdminPassword` with lowercase a instead"
```
jCOTINEAU, line 53: i dont understand the "lowercase a" sentence.

charts/sonarqube-dce/values.schema.json
```diff
@@ -48,15 +48,47 @@
...
+        },
+        "setAdminPassword": {
             "type": "object",
             "properties": {
+                "adminPassword": {
```
jCOTINEAU, line 58: Here conceptually its weird to me as we are adding `setAdminPassword.adminPassword`, its a new set of properties somehow, and we directly mark it as deprecated.

While I understand that due to our merge logic, those sub-keys are still valid and should not be used, its seems a bit strange.

jCOTINEAU (commented):
Should we modify the tests `change-admin-password-hook-values.yaml` and put there the entire new structure? i would expect no changes from the fixture point of view, which would indicate a successful transfer of logic ? maybe also keeping a test file with the legacy `account`?

Wdyt ?

charts/sonarqube-dce/values.schema.json
carminevassallo: Indeed, this is a copy-and-paste issue...

charts/sonarqube-dce/values.schema.json
carminevassallo: > Here conceptually its weird to me as we are adding setAdminPassword.adminPassword, its a new set of properties somehow, and we directly mark it as deprecated.

We are deprecating `adminPassword`, this is not part of what we will suggest to use in the README (although the template logic will allow `setAdminPassword.adminPassword`). In this case, the new value will be `setAdminPassword.newPassword`.

jCOTINEAU: hmm what i mean is that from a yaml point of view `setAdminPassword.adminPassword` never existed before in our value file, we are introducing it right now, hence its strange from a user point of view to see something "new" and deprecated.

While I understand technically its interesting to highlight this property works but should not exist.

This is fine by me I don't have a strong opinion

carminevassallo: I see. So, is your suggestion to remove the deprecation note (in the schema) about `adminPassword`. Because the proper way would be the following, but we can't use it. (Forgot to mention that this is the reason why I did not use it)

```
"account": {
            "type": "object",
            "properties": {
                "adminPassword": {
```

jCOTINEAU (approved):
all good to me.

charts/sonarqube-dce/values.schema.json
carminevassallo: We solve the issue by adding nested deprecations on the "account". Change submitted.

PR #598, opened 2024-12-03
REL-3428 REL-3427 Release SonarQube Server 10.8, SonarQube Community Build 24.12

davividal:


charts/sonarqube/README.md
```diff
@@ -255,7 +255,7 @@ The following table lists the configurable parameters of the SonarQube chart and
 | `annotations`           | SonarQube Pod annotations                                                                                                               | `{}`               |
 | `edition`               | SonarQube Edition to use (e.g. `community`, `developer` or `enterprise`). Please note that the default `community` value is deprecated. | `community`        |
 | `community.enabled`     | Install SonarQube Community Build. When set to `true`, this parameter replaces `edition=community`                                      | `true`             |
-| `community.buildNumber` | The SonarQube Community Build number to install                                                                                         | `10.7.0`           |
+| `community.buildNumber` | The SonarQube Community Build number to install                                                                                         | `10.8.0`           |
```
jCOTINEAU, line 258: community.buildNumber should not be 10.8.0 ?

jCOTINEAU (changes requested):

charts/sonarqube/README.md
davividal: Nice catch!

jCOTINEAU (approved):
lgtm.

sichapman:
This has broken community edition deployment.

Initially get image pull error "10.8.0-community", I think because of https://github.com/SonarSource/helm-chart-sonarqube/blob/master/charts/sonarqube/templates/_helpers.tpl#L65 causing it to skip the build number logic because `image.tag` is set by default.

Tried to work around it by setting `image.tag: ""` but that also fails due to it then changing label `app.kubernetes.io/version: ""` which is immutable.

Had to uninstall everything to get around that, guessing that was not the intention...

jCOTINEAU:
Thanks a lot for reporting this, we will investigate promptly and come back with updates.

verdel:
This issue is not caused by changes in the Helm chart but rather by the fact that the Docker Hub currently does not have an image published with the tag `10.8.0-community`.

The reason the Docker image was not published during the release is this [PR](https://github.com/SonarSource/docker-sonarqube/pull/715) in the [SonarSource/docker-sonarqube](https://github.com/SonarSource/docker-sonarqube) repository. In the GitHub Actions code, the version `10-community` was mistakenly [omitted](https://github.com/SonarSource/docker-sonarqube/blob/6e913e1f2c27f2a6c5eeded46ffaa08f7a940ec6/.github/workflows/release.yml#L37-L40).

verdel:
This issue is not caused by changes in the Helm chart but rather by the fact that the Docker Hub currently does not have an image published with the tag `10.8.0-community`.

jCOTINEAU:
Hello, everyone; I still don't have an ETA for the fixes. In the meantime, as pointed out by @verdel, you can override `image.tag` with `24.12.0.100206-community`.

For the other bug, the workaround is to delete the deployment/statefulset before doing the upgrade, this will prevent the patch of immutable fields and will create things from scratch.

Please note that this might be required again when upgrading to 10.8.1+ if we revert the commit.

PR #606, opened 2024-12-13
SONAR-23559 Improves editions and versions setting for sonarqube chart

davividal:
[SONAR-23559](https://sonarsource.atlassian.net/browse/SONAR-23559)

[SONAR-23559]: https://sonarsource.atlassian.net/browse/SONAR-23559?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.cirrus/schema_test.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+set -euo pipefail
+
+cd "$(dirname "$0")/../tests/dynamic-compatibility-test"
+
+go mod tidy
+go test -timeout=0 -v schema_test.go
```
carminevassallo, line 8: It's super cool that you added more tests! It will be then easier to spot the existence of drift from our desired tagging logic! However, I consider this as static tests, as they do not require to apply our helm chart to a cluster and run SonarQube. Can we move schema_test.go to the `unit-compatibility-test` folder?

charts/sonarqube/values.schema.json
```diff
@@ -205,5 +200,68 @@
                 }
             }
         }
-    }
- }
+    },
+    "allOf": [
```
carminevassallo, line 204: Based on the acceptance criteria of the task, I think that adding this strong validation of the fields is not required.

Specifically, when edition is set, as of now we always favor this, regardless of whether community.enabled is set or not. The downside of adding the strong validation is to have quite a few warnings (not really meaningful) written when edition is not set. 

E.g.
```
Error: values don't meet the specifications of the schema(s) in the following chart(s):
sonarqube:
- (root): Must validate "then" as "if" was valid
- (root): edition is required
- (root): Must validate all the schemas (allOf)
```
Only the second warning is really actionable. 

Having said that, shall we simplify the validation? wdyt?

carminevassallo (changes requested):
Hi @davividal,

the basic logic works, I could validate locally 🎉 

However, there are a few things that I'd suggest we improve. Please let me know your opinion.

.cirrus/schema_test.sh
davividal: Partially agree.
I think that the `unit-compatibility-test` tests are dependant on Kubernetes versions, while these new tests aren't/shouldn't.

Not only that, but the fixtures that we use for other tests are stored within the `unit-compatibility-test`, therefore putting these new tests there could lead to more confusion on the long run.

charts/sonarqube/values.schema.json
davividal: Using schema validation and `fail` will lead to a fail-fast mechanism. Still, I think that `fail` will clutter the templates by trying to mimic something already available to us through schema validation. :)

As for simplifying the validation, when I was rebasing my PR against master (after merging yours), I (expectedly) got the validation error when running the tests. Despite reviewing your PR, it still took me a couple of minutes to understand and locate the template file. Comparing both solutions side-by-side, I don't think that replacing this with that would improve the clarity.

While I agree that the Helm's output is not ideal, I still think the overall solution is easier to understand and maintain.

WDYT?

.cirrus/schema_test.sh
carminevassallo: I like the solution of moving them to the `unit-test` folder (I think it was in the dynamic-compatibility-test` before).

charts/sonarqube/values.schema.json
carminevassallo: Thanks for clarifying. In my opinion, we should favor the readability of the error above all. While I see that in helm we don't really have standard way to fail-fast, the following two error lines that are shown in the current implementation are "internal" errors and should not be shown.

```
- (root): Must validate "then" as "if" was valid
- (root): Must validate all the schemas (allOf)
```
If we follow the "simplified validation", we cleanup the error messages displaying only what's relevant and we don't lose in debugging "power" as there is a clear execution error message pointing to the file, e.g., "Error: execution error at (sonarqube/templates/validation.yaml:2:4)". Furthermore, at the validation-failure stage, the user is not supposed to navigate internally into the chart and debug the failure. If we adopt clear failure messages, the users will be clearly notified about what's wrong and how the deployment can be fixed (e.g., which parameters are not accepted, which ones need to be now provided). My proposal would still be to use the validation through the validation.yaml file, wdyt?

davividal: What do you think now, @carminevassallo?

I kept the enum in the `values.schema`, as the error message seems clear and relevant enough to me:

```
$ helm template charts/sonarqube -f tests/unit-test/test-cases-values/sonarqube/invalid-edition.yaml --set monitoringPasscode=abc42
Error: values don't meet the specifications of the schema(s) in the following chart(s):
sonarqube:
- edition: edition must be one of the following: "developer", "enterprise"
```

carminevassallo: Hi @davividal,

just a few comments on the latest changes :)


1. Move the whole logic about edition setting in validation.yaml

I suggest that we keep the full fail fast mechanism in the `validation.yaml`, otherwise we have different messages printed.

Ex.
When using this values file:

```
monitoringPasscode: "test"
```

We get:

```
Error: execution error at (sonarqube/templates/validation.yaml:22:4): 
 ** The values.yaml file is not valid. ** 
 You must choose an 'edition' to install: 'developer' or 'enterprise'.
```

If we make a mistake and we use:
```
monitoringPasscode: "test"
edition: "developer"
```

we get:

```
Error: values don't meet the specifications of the schema(s) in the following chart(s):
sonarqube:
- edition: edition must be one of the following: "developer", "enterprise"
```

We should move the whole logic to the validation.yaml also because in case of `edition=community` we don't just want to print out that this is not valid but also suggest to use `community.enabled=true`.

2. Rephrase message in case of edition=community

This message is slightly wrong. Users can either set edition (for paid editions) or community.enabled=true (for community build).

```
** The values.yaml file is not valid. ** 
 You must choose an 'edition' to install: 'developer' or 'enterprise'.

File: charts/sonarqube/templates/validation.yaml [hunk 3]
```diff
@@ -1,3 +1,29 @@
...
+{{/*
+* Validates the community.enabled, edition, and tag logic.
+*/}}
+{{- if eq .Values.edition "community" -}}
+    {{- include "sonarqube.fail" "'community' is not a valid edition. If you want to use SonarQube Community Build, unset 'edition' and set 'community.enabled=true' instead." -}}
+{{- else if and (.Values.community.enabled) (not (empty .Values.edition)) -}}
+    {{- include "sonarqube.fail" "You can't set 'community.enabled=true' and an 'edition' at the same time." -}}
+{{- else if not .Values.community.enabled -}}
+    {{- if empty .Values.edition -}}
+        {{- include "sonarqube.fail" "You must choose an 'edition' to install: 'developer' or 'enterprise'.\nIf you want to use SonarQube Community Build, unset 'edition' and set 'community.enabled=true' instead." -}}
```
carminevassallo, line 25: The message printed is 
image

I'd like that we remove the misalignment of the second message. I'd suggest we remove the "\n" as the whole message does not see too long IMO.

charts/sonarqube/README.md
```diff
@@ -240,29 +239,29 @@ The following table lists the configurable parameters of the SonarQube chart and
...
+| `tolerations`           | List of node taints to tolerate                                                                                       | `[]`               |
+| `nodeSelector`          | Node labels for pod assignment                                                                                        | `{}`               |
+| `hostAliases`           | Aliases for IPs in /etc/hosts                                                                                         | `[]`               |
+| `podLabels`             | Map of labels to add to the pods                                                                                      | `{}`               |
+| `env`                   | Environment variables to attach to the pods                                                                           | `{}`               |
+| `annotations`           | SonarQube Pod annotations                                                                                             | `{}`               |
+| `edition`               | SonarQube Edition to use (`developer` or `enterprise`).                                                               | `None`             |
```
carminevassallo, line 256: Personal opinion (Feel free to ignore my comment if you disagree). Shall we add "If you want to use SonarQube Community Build, set 'community.enabled=true' instead."" here? It might be less disruptive of people using community understand from the parameter description what to use.

carminevassallo (approved):
LGTM! Just two small comments that are not blocking the acceptance.

PR #616, opened 2025-01-08
SONAR-24105 remove 9.9 deprecated feature

jCOTINEAU:
[SONAR-24105](https://sonarsource.atlassian.net/browse/SONAR-24105)

Part of 

[SONAR-24105]: https://sonarsource.atlassian.net/browse/SONAR-24105?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


carminevassallo (changes requested):
Hey @jCOTINEAU, 

I think we can delete more :) If I look at the latest [sonarqube LTS](https://artifacthub.io/packages/helm/sonarqube/sonarqube/8.0.8+3946), we have also `image.pullSecret` and `postgresql.postgresqlServer` to be removed. On the latest [sonarqube-dce LTS](https://artifacthub.io/packages/helm/sonarqube/sonarqube-dce/7.0.8+3946) we have `image.pullSecret` on both application and search nodes.

jCOTINEAU:
indeed, i completely overlooked those, completely missed the search, might have made a wrong checkout.

tests/unit-compatibility-test/fixtures/sonarqube-dce/secret-values.yaml
```diff
@@ -223,7 +223,7 @@ metadata:
     heritage: Helm
 data:
   SONAR_JDBC_USERNAME: sonarUser
-  SONAR_JDBC_URL: "jdbc:postgresql://%!s(<nil>):5432/sonarDB"
+  SONAR_JDBC_URL: "jdbc:postgresql://:5432/sonarDB"
```
jCOTINEAU, line 226: here this highlight a problem we had before, in this test case we have no jdbcoverwrite as well as postgresql.enabled=false.

Hence the default data in the configmap does not make sense.

i am creating a hardening ticket to complete the migration to having config map, secret and so on, created only when it make sense.

Here for example the configmap should exist only if jdbcoverwrite or postgresql.enabled, and should fail fast in dce

PR #605, opened 2024-12-11
Add the possibility of to save the data with hostpath.

jbazp:
Sonarqube will not have more than one instance at a time, I want it to always run on a node and within that node save the information with hostpath, I do not want PVC or PV, I want to directly access the host disk to save it and it will go very quickly , easy and without intermediaries (Storage Provider).

Please ensure your pull request adheres to the following guidelines:
- [x] Document your Changes in the `CHANGELOG.md` file of the respected chart as well as the `Chart.yaml`

PR #631, opened 2025-02-03
[Do not merge] Add the possibility of to save the data with hostpath

davividal:
This duplicates #605 so we can trigger the CI


charts/sonarqube-dce/templates/sonarqube-application.yaml
```diff
@@ -490,7 +490,13 @@ spec:
           secretName: {{ .Values.searchNodes.searchAuthentication.keyStoreSecret }}
       {{- end }}
       - name: sonarqube
+        {{- if .Values.hostPath.enabled }}
+        hostPath:
```
carminevassallo, line 494: We don't persist the application nodes's storage in persistent volumes. Why do we want to add the hostPath here?

charts/sonarqube/README.md
```diff
@@ -482,19 +482,22 @@ The following table lists the configurable parameters of the SonarQube chart and
...
+| `persistence.existingClaim` | Do not create a new PVC but use this one            | `None`          |
+| `persistence.storageClass`  | Storage class to be used                            | `""`            |
+| `persistence.accessMode`    | Volumes access mode to be set                       | `ReadWriteOnce` |
+| `persistence.size`          | Size of the volume                                  | `5Gi`           |
+| `persistence.volumes`       | (DEPRECATED) Please use extraVolumes instead        | `[]`            |
+| `persistence.mounts`        | (DEPRECATED) Please use extraVolumeMounts instead   | `[]`            |
+| `persistence.uid`           | UID used for init-fs container                      | `1000`          |
+| `persistence.guid`          | GUID used for init-fs container                     | `0`             |
+| `emptyDir`                  | Configuration of resources for `emptyDir`           | `{}`            |
+| `hostPath.enabled`          | Flag for enabling storage with `hostPath`           | `false`         |
```
carminevassallo, line 498: should this be part of the `persistence` value? It's all about persistency at the end. Users can decide where to use a persistent volume or leverage the hostpath. WDYT?

carminevassallo (changes requested):
Tested on the standard chart and works as expected. However, I do have a few comments that might require some changes...

carminevassallo (approved):

PR #645, opened 2025-04-01
chore: use the quote function to handle the .Release.Name

kycheng:
When deploying via helm, an error is reported when the helm Release.Name is a numeric, because the label cannot use numbers. 

image

This problem can be solved by the handling of the `quote` function.

### DIFF
--- a/charts/sonarqube-dce/templates/_helpers.tpl
+++ b/charts/sonarqube-dce/templates/_helpers.tpl
@@ -73,7 +73,7 @@ Common labels
 {{- define "sonarqube.labels" -}}
 app: {{ include "sonarqube.name" . }}
 chart: {{ printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" }}
-release: {{ .Release.Name }}
+release: {{ .Release.Name | quote }}
 heritage: {{ .Release.Service }}
 {{- end -}}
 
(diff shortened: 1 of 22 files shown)


kycheng:
@davividal hello, Can you help review this PR? I think this modification is valuable.

jCOTINEAU:
Hello @kycheng, thanks a lot for participating in the community, according to kubernetes label specification :

```
The name segment is required and must be 63 characters or less, beginning and ending with an alphanumeric character ([a-z0-9A-Z]) with dashes (-), underscores (_), dots (.), and alphanumerics between. The prefix is optional. If specified, the prefix must be a DNS subdomain: a series of DNS labels separated by dots (.), not longer than 253 characters in total, followed by a slash (/).
```

Labels and other kubernetes field can never be alphanumerical only, is there a particular reason why you want to achieve this ?

kycheng:
@jCOTINEAU There are also problems with deploying some Boolean keywords, although this is not a problem with deploying sonar-chart. 
image
image
but I think it can be avoided in this way.

jCOTINEAU:
Ha yes, forget everything, we will take a deeper look, can you tell me weather its for buisness or personal ?

kycheng:
Our company encountered this problem when using the chart and modified it by forking it.

kycheng:
@jCOTINEAU Whether or not to consider accepting this change?

jCOTINEAU:
Hello @kycheng, sorry i think there is not enough value added here, as we have not received any hints for naming a release with booleans as it does not follow kubernetes RFC

PR #651, opened 2025-04-11
Use exec command from liveness probe for startup probe by default

florianmutter:
The purpose of the startup probe is to tell kubernetes when an application is started. As soon as the probe is successful kubernetes stops checking it and starts checking the liveness and readiness probe. See https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#types-of-probe

With the current configuration it sometimes happens that the startup probe is successful but the subsequent liveness checks are not and kubernetes kills the pod. By aligning the startup probe with the liveness probe, it should avoid these situations and allow the sonar pods to start and run smoothly.


vnaipaul:
Hi @florianmutter ,
Setting aside the proposed solution for the moment, have you seen the response body from the startup probe (`/api/system/status`) on these occasions ?

Is the startup probe succeeding (as an HTTP probe), but masking a degraded state--its [`api/system/status` endpoint is documented as returning](https://next.sonarqube.com/sonarqube/web_api/api/system/status?internal=true) "STARTING", "RESTARTING" or even "DOWN"--which fails the liveness probe, afterwards ?

Fwiw I just converted a readiness probe using `api/system/status` (i.e. the default) to `api/system/health`, on a hunch that it's the more future-proof option.

florianmutter:
Hey we have not checked the output we only observed that startup was successful but liveness was not. I think both should use the same logic to determine success or failure. Which one it is I cannot say :) for this do not know enough about the system status of sonar.

vnaipaul:
Signs of a dilemma :) Keen to hear what the Sonarqubists think.

jCOTINEAU:
Hello everyone, thanks again for your time and patience helping us grow our product trough community.

You can track the proposal in this [ticket](https://sonarsource.atlassian.net/browse/SONAR-25087), i am closing the PR for now, if we reopen, your commit will be used and you will be acknowledged as author.

We tackle community PR in batch during hardening sprints, i dont have yet an ETA for the feature.

I cannot yet answer definitely the topic questions, but can already state that yes, the different endpoints are important, as sometimes (for example during upgrades) only a sub-part of sonarqube will be up-and running, and this needs to be detected by the probes.

Cheers, Jeremy.

PR #657, opened 2025-04-22
Allow JDBC URL to be read from a secret

tobimichael96:
Hi all,

we are creating our databases via a Terraform module which creates the DB and the role for it. It also creates and AWS secret containing all needed information like `role-name`, `role-password` and `database-url`. 

Currently, we have to hardcode the URL of the database to the helm chart values. We would like to avoid that and grab the URL directly from the already existing secret (where we also get the password from). 

This change should make that possible. I tried to integrate it into the current logic without changing too much and also make it backwards compatible. 

Let me know what you think!

___

### DIFF
--- a/charts/sonarqube/values.yaml
+++ b/charts/sonarqube/values.yaml
@@ -529,7 +529,7 @@ jdbcOverwrite:
   ## and the secretValueKey of the password found within that secret
   # jdbcSecretPasswordKey: "jdbc-password"
   ## and the secretValueKey of the URL found within that secret
-  # jdbcSecretURLKey: "jdbc-password"
+  # jdbcSecretURLKey: "jdbc-url"
   # To install the oracle JDBC driver, set the following URL (in this example, we set the URL for the Oracle 11 driver. Please update it to your target driver URL.).
   # If downloading the driver requires authentication, please set the .netrc secret file with a key "netrc" to use basic auth.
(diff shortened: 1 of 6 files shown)


lchastel:
I need this feature for the same reasons

charts/sonarqube/values.yaml
```diff
@@ -528,6 +528,8 @@ jdbcOverwrite:
   # jdbcSecretName: "sonarqube-jdbc"
   ## and the secretValueKey of the password found within that secret
   # jdbcSecretPasswordKey: "jdbc-password"
+  ## and the secretValueKey of the URL found within that secret
+  # jdbcSecretURLKey: "jdbc-password"
```
lchastel, line 532: Please change the value to "jdbc-url" instead of "jdbc-password"

tobimichael96:
Any more feedback on this?

jCOTINEAU:
Hello @tobimichael96, thanks again for your time and patience helping us grow our product trough community.

You can track your proposal in this [ticket](https://sonarsource.atlassian.net/browse/SONAR-25086), i am closing the PR for now, if we reopen your commit will be used and you will be acknowledged as author.

We tackle community PR in batch during hardening sprints, i dont have yet an ETA for the feature.

Cheers, Jeremy.

PR #669, opened 2025-05-28
SONAR-25081 release 2025.1.2

jCOTINEAU:


carminevassallo (changes requested):
Thanks for addressing the changes!

Unfortunately, while looking at the current tests, we have at least one thing that require immediate action: we are using outdated images in the cirrus tests (`ci/cirrus-values.yaml`). We should update them before releasing (for the standard chart I suggest we test an enterprise edition instead of community).

carminevassallo (approved):
LGTM!

PR #705, opened 2025-08-07
SONAR-25633 Trigger SonarQube 2025.5.0 release cycle

carminevassallo:


carminevassallo:
> I think we are missing update on sonarqube-dce/README.md as well as the `GCLOUD_TAG` from tasks_env.yml ?

We do those updates at the release time, don't we?

jCOTINEAU (approved):
> > I think we are missing update on sonarqube-dce/README.md as well as the `GCLOUD_TAG` from tasks_env.yml ?
> 
> We do those updates at the release time, don't we?

HA, for the README you are completly right it would not make sense, got confused with community build.

And for the GCLOUD version, right now if we dont change that mean staging helm chart will use previous images , which is probably all good conceptually.

My bad, everything is correct, approving

PR #731, opened 2025-09-30
SONAR-25923 Set azure clientId, planId, and resourceId as env variables

carminevassallo:


carminevassallo:
> i see the configMap with the ENV, but i dont see that in the deployment.yaml, is that expected ?

Isn't sonarqube.pod included also in the deployment.yaml?

jCOTINEAU:
> > i see the configMap with the ENV, but i dont see that in the deployment.yaml, is that expected ?
> 
> Isn't sonarqube.pod included also in the deployment.yaml?

hooo yes, my bad overlooked. all good

PR #785, opened 2026-01-30
fix: add option to disable proxy configuration

afnanashraf10:
## Description

Hey there! 👋

I ran into an issue where the plugin install init container would fail if no proxy was configured. The current setup always tries to set proxy environment variables, even when they're empty strings, which causes problems in environments without a proxy.

## Changes

Added a `proxy.enabled` flag to control whether proxy settings should be applied at all. This prevents empty proxy values from being passed to the init containers when you're not using a proxy.

**Before:**
```yaml
httpProxySecret: ""
httpProxy: ""
httpsProxy: ""
noProxy: ""
```

**After:**
```yaml
proxy:
  enabled: false
  httpProxySecret: ""
  httpProxy: ""
  httpsProxy: ""
  noProxy: ""
```

## Why this matters

Without this change, if you don't have a proxy in your environment, the init containers get empty proxy env vars which can cause them to fail or behave unexpectedly. Now you can explicitly disable proxy configuration, and those env vars won't be set at all.

## Backwards compatibility

This is backwards compatible - if someone was already setting proxy values, they just need to add `enabled: true` to keep the same behavior.

Let me know if you'd like me to adjust anything!


jimil09 (changes requested):
Hi @afnanashraf10! :)

Thanks for your contribution. I had a look and you're correct that empty proxy environment variables can cause download failures in the init containers. Before I merge, I have some concern I would like to point out:

### Breaking Change Concern

The nested structure introduces a breaking change for existing users:

**Current configuration (works today):**
```yaml
httpProxy: "http://proxy.corp.com:8080"
```

**After this PR (requires migration):**
```yaml
proxy:
  enabled: true
  httpProxy: "http://proxy.corp.com:8080"
```

Users would need to add `enabled: true` and restructure their values, which could silently break deployments during upgrades.

Instead I suggest an alternative approach:

### Alternative Approach: Conditional Injection

Instead of requiring an `enabled` flag and nested structure, we can conditionally inject the proxy environment variables only when values are actually set. This keeps the flat structure and is truly backwards compatible.

**Implementation:**

In `charts/sonarqube/templates/_helpers.tpl` and `charts/sonarqube-dce/templates/_helpers.tpl`, we could wrap the helper definitions with a conditional check (in `_helpers.tpl`):

```go
{{- define "sonarqube.prometheusExporterProxy.env" -}}
{{- if or .Values.httpProxySecret .Values.httpProxy .Values.httpsProxy .Values.prometheusExporter.httpProxy .Values.prometheusExporter.httpsProxy -}}
{{- if .Values.httpProxySecret -}}
{{- include "sonarqube.proxyFromSecret" . }}
{{- else -}}
- name: http_proxy
  valueFrom:
    secretKeyRef:
      name: {{ template "sonarqube.fullname" . }}-http-proxies
      key: PROMETHEUS-EXPORTER-HTTP-PROXY
- name: https_proxy
  valueFrom:
    secretKeyRef:
      name: {{ template "sonarqube.fullname" . }}-http-proxies
      key: PROMETHEUS-EXPORTER-HTTPS-PROXY
- name: no_proxy
  valueFrom:
    secretKeyRef:
      name: {{ template "sonarqube.fullname" . }}-http-proxies
      key: PROMETHEUS-EXPORTER-NO-PROXY
{{- end -}}
{{- end -}}
{{- end -}}
```

Apply the same pattern to `sonarqube.install-plugins-proxy.env`.

**Result:**
- Empty proxy values = no environment variables injected (solves your issue)
- Existing configurations continue to work without modification
- No additional `enabled` flag to manage
- Follows Helm conventions where value presence drives behavior

## Additional Issues

There are typos in the test files where `enbled` should be `enabled`:
- `tests/unit-compatibility-test/sonarqube-dce/proxies-values.yaml` lines 469, 478
- `tests/unit-compatibility-test/sonarqube/proxies-values.yaml` lines 535, 544, 552

---

Would you be willing to adjust the approach to avoid the breaking change? Let me know if you'd like clarification on the implementation.

afnanashraf10:
Hi @jimil09! Thanks for the detailed review and suggestions.

I've made all the requested changes:

**Breaking Change Fix**
Reverted the nested `proxy` structure back to the flat configuration, so existing users won't need to migrate their values. Proxy environment variables are now conditionally injected only when values are actually set, keeping full backwards compatibility.

**Implementation Note**
Rather than wrapping the conditionals in `_helpers.tpl` as suggested, I've applied the condition checks in `charts/sonarqube/templates/_pod.tpl` and `charts/sonarqube-dce/templates/sonarqube-application.yaml` instead, so the env block is skipped at the template level when no proxy values are present.

**Typo Fixes**
Removed proxy.enabled in:
- `tests/unit-compatibility-test/sonarqube-dce/proxies-values.yaml`
- `tests/unit-compatibility-test/sonarqube/proxies-values.yaml`

Let me know if anything else needs adjusting!

tests/unit-compatibility-test/sonarqube-dce/duplicated-env-values.yaml
```diff
@@ -14,9 +14,11 @@ ApplicationNodes:
 
 prometheusExporter:
     enabled: true
-    httpProxy: "http://someProxy"
-    httpsProxy: "https://someProxy"
-    noProxy: "https://someUrl"
+    proxy:
+      enabled: true
```
jimil09: This should be dropped since we now use a flat structure :)

tests/unit-compatibility-test/sonarqube-dce/install-plugins-values.yaml
```diff
@@ -3,6 +3,8 @@ ApplicationNodes:
     noCheckCertificate: true
     install:
       - "https://github.com/sbaudoin/sonar-yaml/releases/download/v1.7.0/sonar-yaml-plugin-1.7.0.jar"
-    httpProxy: "http://someProxy"
-    httpsProxy: "https://someProxy"
-    noProxy: "https://someUrl"
\ No newline at end of file
+    proxy:
+      enabled: true
```
jimil09: same as above; we can drop this :)

tests/unit-compatibility-test/sonarqube-dce/prometheus-monitoring-values.yaml
```diff
@@ -2,6 +2,8 @@ ApplicationNodes:
   prometheusExporter:
     enabled: true
     noCheckCertificate: true
+    proxy:
+      enabled: true
```
jimil09: same as above!

jimil09 (changes requested):
Thanks for making those changes! The backwards compatibility fix looks good, and I appreciate you keeping the flat structure.

I noticed a few issues with the test files that need to be addressed before merge:

### Test Files Still Use Nested Structure

The test files are using a nested `proxy.enabled` (check the review below) structure that doesn't match the actual implementation:

**In `tests/unit-compatibility-test/sonarqube/duplicated-env-values.yaml`:**
```yaml
prometheusExporter:
    enabled: true
    proxy:
      enabled: true
      httpProxy: "http://someProxy"
      httpsProxy: "https://someProxy"
      noProxy: "https://someUrl"
```

**Should be:**
```yaml
prometheusExporter:
    enabled: true
    httpProxy: "http://someProxy"
    httpsProxy: "https://someProxy"
    noProxy: "https://someUrl"
```

The same issue exists in:
- `tests/unit-compatibility-test/sonarqube-dce/duplicated-env-values.yaml`
- `tests/unit-compatibility-test/sonarqube-dce/install-plugins-values.yaml`
- `tests/unit-compatibility-test/sonarqube-dce/prometheus-monitoring-values.yaml`

These need to use the flat structure to match the actual values.yaml schema.

### Implementation Note

The current approach of adding conditionals in the pod templates works. However, for consistency and to avoid repetition, we could alternatively wrap the helper template definitions in `_helpers.tpl` instead (as suggested before :D). That way the condition logic lives in one place rather than being duplicated across multiple pod template files. Either approach works, but the helper-based approach would be slightly cleaner.

---

Could you update the test files to use the flat structure consistently? Let me know if you have any questions.

afnanashraf10:
Hi @jimil09! Thanks for pointing those out! I've updated all the test files to use the flat structure. Let me know if there's anything else to fix!

afnanashraf10:
@jimil09 Let me know if any further changes are required.

hatem-amairi-sonarsource:
Hello @afnanashraf10
Thanks again for the contribution, and sorry again for the delay.

I rebased the branch on top of the current master and pushed a follow-up fix for the partial proxy configuration edge case, along with the regenerated fixtures and chart version/changelog updates. I hope that's ok for you. 

The edge case issue was that the proxy Secret and proxy env vars could still be rendered as a group when only part of the proxy configuration was set.

Sent for a final review to @jimil09

jimil09 (approved):

afnanashraf10:
@jimil09 can we merge this PR. Is there anything else required. Please let me know.

PR #794, opened 2026-02-16
Support annotations in HTTPRoute

Brawdunoir:
# Motivation

Most external tools (like external-dns or cert-manager) rely on annotations instead of labels. Currently, only labels can be set while using the chart, which limit the interaction with external tools.


jimil09 (changes requested):
Hey @Brawdunoir! 👋 

Thanks for submitting this PR; the feature makes sense and the implementation looks good! I have a few suggestions to make this complete:

### 1. `sonarqube-dce` Chart

The `sonarqube-dce` (Data Center Edition) chart also has HTTPRoute support but wasn't updated in this PR. Since both charts should maintain feature parity, could you please add the same annotation support to:

- `charts/sonarqube-dce/templates/http-route.yaml` (add annotations block)
- `charts/sonarqube-dce/values.yaml` (add commented example)
- `charts/sonarqube-dce/CHANGELOG.md` (add entry)

### 2. CHANGELOG Version

The CHANGELOG entry was added to `[2026.2.0]`, but since that version has already been released, this should be moved to the upcoming `[2026.3.0]` section

### 3. Chart.yaml Metadata

Could you also add an entry to the `artifacthub.io/changes` annotation in both `charts/sonarqube/Chart.yaml` and `charts/sonarqube-dce/Chart.yaml`? Something like:

```yaml
artifacthub.io/changes: |
  - kind: added
    description: "Support annotations in HTTPRoute"

Brawdunoir:
Hello @jimil09, sorry for the delay. Done !

PR #808, opened 2026-03-10
SONAR-27253 Migrate CI tasks to GH Actions (release/2025.4)

hatem-amairi-sonarsource:
Migrate CI tasks to GH Actions

(cherry picked from commit 9a06b87e2aaa3fc82ca3861b2ba4123940bdc36d)


carminevassallo (changes requested):
Hi @hatem-amairi-sonarsource,

great job on the backporting 👍  However, we have one major thing to still add here. Since `2025.4.0`, we support Istio. So we need to backport (from `master` this time) the whole testing with Istio...

(Minor suggestion) Can we remove the `.github/scripts/azure_app_verify_and_publish.sh` file? (It's not used, we should have removed on 2025 too :P)

carminevassallo (approved):
LGTM!

PR #820, opened 2026-03-30
SONAR-27394 Add support for TopologySpreadConstraint on the sonarqube-dce Helm chart

jimil09:
More about [topologySpreadConstraint](https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/).


charts/sonarqube-dce/Chart.yaml
```diff
@@ -41,6 +41,8 @@ annotations:
       description: "Replace wget with curl in health probes"
     - kind: changed
       description: "Use -fS flag in curl to show errors in liveness and startup probes"
+    - kind: changed
```
carminevassallo: Isn't this more an "added" type? See them -> https://artifacthub.io/docs/topics/annotations/helm/

carminevassallo (commented):
Hey @jimil09,

just minor comments on your PR. I was playing with setting hostname and zone TSC locally, the templates are rendered correctly. Basically, it's ready to be shipped 🚀 . 

I have a few considerations to make, mostly for documentation purposes. I think it's good to leave the TSC empty by default, however, we could suggest some best practices for production.

In case of a disaster recovery (e.g., one zone is down), the app pods do not have difficulty to get automatically rescheduled on any other nodes. However, for search pods which are stateful this is trickier right? We should really try to spread them across different zones at the installation/upgrade phase. This is because we should have at least 2 Elasticsearch pods running to form the cluster, so we can't afford losing 2 out of 3 if they are scheduled on the same zone that is down. Plus, given that PVCs are typically bound to nodes, we can't reschedule them on a different zone (and its nodes).

That's why maybe we could add some docs in the README (maybe inside the `Prerequisites and suggested settings for production`) to suggest to spread search pods specifically using the new feature (plus proper node labels). And add some examples like this?

```
topologySpreadConstraints:
    - maxSkew: 1
      topologyKey: topology.kubernetes.io/zone
      whenUnsatisfiable: DoNotSchedule
      labelSelector:
        matchLabels:
          app: sonarqube-dce-search
```

carminevassallo (changes requested):

charts/sonarqube-dce/values.schema.json
```diff
@@ -168,6 +168,9 @@
                     "deprecated": true,
                     "$comment": "PodDisruptionBudget for the Application Nodes"
                 },
+                "topologySpreadConstraints": {
```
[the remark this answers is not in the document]
jimil09: @carminevassallo I added it for early type validation, since Helm does not catch this by itself without schema validation. But if the convention here is to keep values.schema.json mostly for deprecations/minimal checks, I’m fine removing these entries for consistency.

carminevassallo (changes requested):
Nice addition to the docs, @jimil09!

One last change before merging!

charts/sonarqube-dce/Chart.yaml
```diff
@@ -2,8 +2,8 @@ apiVersion: v2
 name: sonarqube-dce
 description: SonarQube is a self-managed, automatic code review tool that systematically helps you deliver clean code. As a core element of our Sonar solution, SonarQube integrates into your existing workflow and detects issues in your code to help you perform continuous code inspections of your projects. The tool analyses 30+ different programming languages and integrates into your CI pipeline and DevOps platform to ensure that your code meets high-quality standards.
 type: application
-version: 2026.2.0
-appVersion: 2026.2.0
+version: 2026.3.0
+appVersion: 2026.3.0
```
carminevassallo: The app version stays `2026.2.0`, as we haven't released the docker images yet.

carminevassallo (changes requested):

carminevassallo (approved):

PR #824, opened 2026-04-02
SONAR-27402 Add clarification note for openshift vs k8s version support

jimil09:


carminevassallo (changes requested):
Good note, @jimil09!

Please feel free to remove the changelog about the edit in the documentation. We don't need to log it.

carminevassallo (approved):

PR #831, opened 2026-04-13
REL-4517: Release SonarQube Community Build 26.4

jimil09:


charts/sonarqube/CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 # SonarQube Chart Changelog
 All changes to this chart will be documented in this file.
 
+## [2026.3.0]
```
carminevassallo, line 4: We should add an entry about the the new chart's version (Upgrade Chart's version to 2026.3.0). I noticed we forgot to do it when bumping the dce chart. Can we add the entry there as well?

charts/sonarqube/Chart.yaml
carminevassallo: The version should be updated here.

charts/sonarqube/Chart.yaml
```diff
@@ -48,10 +48,12 @@ annotations:
       description: "Replace wget with curl in health probes"
     - kind: changed
       description: "Use -fS flag in curl to show errors in liveness probes"
+    - kind: added
```
carminevassallo, line 43: This is good, however, the previous changes should only include those related to 2026.3.0

carminevassallo (changes requested):

carminevassallo (approved):

PR #838, opened 2026-05-12
SONAR-27806 Fix and refactor dynamic compatibility tests

navyatejaotturu-sonarsource:
## Summary

Reworks the SonarQube dynamic compatibility tests under `tests/dynamic-compatibility-test/` so they:
- Connect to an **external PostgreSQL** instead of the soon-to-be-removed embedded DB
- Live in **per-case files** in chart-specific subpackages instead of one large table-driven test
- Wait for chart readiness via `kubectl rollout status` — replacing custom pod-snapshot waits that produced false positives and flaked under cluster pressure
- Run in **parallel** across packages, cutting wall-clock time
- Are launched through a single entry-point script (`run_dynamic_compatibility_tests.sh`) usable both locally and in CI

### Test layout
- `sonarqube_standard_dynamic_test.go` (deleted) is replaced by per-case files split across two chart-specific subpackages:
  - `sonarqube/`: `deployment_with_default_values_test.go`, `deployment_with_all_values_test.go`, `statefulset_with_default_values_test.go`, `statefulset_with_all_values_test.go`
  - `sonarqube-dce/`: `with_default_values_test.go`, `with_all_values_test.go`
- Each test is ~20 lines — chart name, values, values files, and an `RequireExternalDB` flag passed to `utils.RunChartTest`
- Every top-level test calls `t.Parallel()`. The wrapper script runs 2 tests per package × 2 packages concurrently — conservative because each DCE install needs ~17 GB of memory

### Helpers (`tests/dynamic-compatibility-test/`)
- **`test_utils.go`** — `RunChartTest` + `ChartTestSpec`; the per-test boilerplate (namespace, helm options, apply, wait, smoke-test) is now in one place
- **`k8s_utils.go`** —
  - `NamespaceFor(chartName)` derives an RFC 1123-safe namespace name with a random suffix so reruns and concurrent runs never collide
  - `WaitForChartReady(t, opts, chartName)` discovers Deployments and StatefulSets by `release=<chart>` label and runs `kubectl rollout status` on each (15 min timeout). Replaces the custom snapshot-walking waits
  - `CheckSonarQubeUpAndRunning` narrows the label selector for DCE to `sonarqube.datacenter/type=app` so it never tunnels into a search pod (which doesn't listen on 9000)
  - `DeleteNamespaceAndWait` (renamed from the misleading `WaitUntilNamespaceDeletedE`) performs the deletion and short-circuits when the namespace doesn't exist
- **`dependencies/postgres_utils.go`** — `SetupDB(t, opts)` installs Bitnami PostgreSQL 18.2.3 in the test namespace; exposes constants for the JDBC config (`PostgresHost`, `PostgresSecretName`, `PostgresSecretPasswordKey`, …)

### Path resolution
Both the chart path and the Postgres values file path are resolved via `runtime.Caller`, anchoring them to the source file's location rather than the test process's CWD. Earlier the test relied on every caller being exactly 3 directories below the repo root — restructuring or moving the package would have silently broken paths.

### Bug fixes in the waiting logic
The previous `CheckMinimumNumberOfRunningPods` + `CheckPodsInContainersReadyState` + `WaitUntilPodContainersReady*` helpers were deleted because they had four distinct flake sources:
1. The `ContainersReady` condition check returned success when the condition was `Unknown` or absent (false positive)
2. The pod count check accepted `Pending`/`Failed`/`CrashLoopBackOff` pods as "running"
3. The pod-list snapshot went stale if any pod was recreated between snapshot and wait (StatefulSet rolling, OOMKill, eviction)
4. No timeout — the count loop could spin indefinitely

`kubectl rollout status` reads the workload's own `spec.replicas` vs `status.readyReplicas` from the API server in real time, eliminating all four.

### Postgres setup
- Per-test install in the same namespace, torn down with the namespace — no shared DB state between tests
- `dependencies/postgres-values.yaml` hardcodes test credentials (`sonar`/`sonarpass`) and disables persistence to skip the PVC dependency
- `sonarqube-dce/all-values.yaml` references the auto-generated Bitnami secret via `jdbcSecretName` + `jdbcSecretPasswordKey` (the deprecated `jdbcPassword` field is no longer used, per the chart's JSON-schema deprecation rule)
- DCE NetworkPolicy is re-enabled with an `additionalNetworkPolicies` block that allows egress on TCP 5432 to the Postgres pod — the chart's defaults don't include DB egress, so without this the SonarQube → Postgres traffic gets silently dropped

### Infrastructure
- `.github/scripts/run_dynamic_compatibility_tests.sh` — single entry-point. Builds chart deps, then runs `go test -timeout=90m -parallel 2 -p 2 ./...`
- `.github/scripts/build_chart_dependencies.sh` — dropped the obsolete `bitnami-pre2022` repo add (no longer in any `Chart.yaml`)
- `tests/dynamic-compatibility-test/go.mod` — module path corrected from `github.com/gruntwork-io/terratest-helm-testing-example/test` to `github.com/helm-chart-sonarqube/tests/dynamic-compatibility-test`

### Documentation
- README updated to reflect the new entry-point script and Go version (1.25)

## How to validate

Prereqs: Kubernetes cluster sized for DCE (~22 GB memory), Go 1.25, Helm, kubectl.

```bash
./.github/scripts/run_dynamic_compatibility_tests.sh

### CODE REVIEW

File: tests/dynamic-compatibility-test/k8s_utils.go [hunk 1]
```diff
@@ -0,0 +1,133 @@
...
+		fmt.Printf("Deleting an existing namespace: %v\n", namespaceActualName)
+		k8s.DeleteNamespaceE(t, options, namespaceActualName) // asynchronous call
+		// verify if the namespace still exists, it takes some time to delete it
+		namespace, err = k8s.GetNamespaceE(t, options, namespaceName)
+		if err != nil || namespace == nil {
+			return //namespace doesn't exist, nothing to delete
+		}
+
+		namespaceActualName = namespace.Name
+		time.Sleep(10 * time.Second)
```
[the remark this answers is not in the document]
navyatejaotturu-sonarsource: I see it is guarded, there is "if" checking if namespace is nil

tests/dynamic-compatibility-test/dependencies/postgres-values.yaml
```diff
@@ -0,0 +1,24 @@
+# Bitnami PostgreSQL values for SonarQube dynamic compatibility tests.
+# Hardcoded credentials — these are test-only and never reach a real environment.
+#
+# The same values must be referenced by the JDBC config passed to the
+# sonarqube-dce chart.
+
+auth:
```
carminevassallo: We should avoid setting those credentials explicitly (even if we use it for testing purposes). From my previous task about setting up an external DB used by DCE in the tests, I remember that the postgresql chart can create a default secret in the namespace. This is the secret we should reference in our test when setting the jdbc url in SonarQube (no need to set those credentials)

tests/dynamic-compatibility-test/sonarqube/deployment_with_all_values_test.go
```diff
@@ -0,0 +1,22 @@
...
+	t.Parallel()
+
+	utils.RunChartTest(t, utils.ChartTestSpec{
+		Name: "sonarqube",
+		Values: map[string]string{
+			utils.TESTS_ENABLING_ACTION:  "false",
+			"deploymentType":             "Deployment",
+			"prometheusExporter.enabled": "false", // only for deployment
+		},
+		ValuesFiles:       []string{"all-values.yaml"},
```
carminevassallo, line 19: Where is such a all-values file? Wasn't checked in to the branch?

carminevassallo (changes requested):
Great job with this refactoring, @navyatejaotturu-sonarsource!

Also great you fixed the bug with the external database. I left a few comments for you...the current tests with the current values are running correctly on my local cluster..

tests/dynamic-compatibility-test/sonarqube/deployment_with_all_values_test.go
navyatejaotturu-sonarsource: The `all-values.yaml` is present under the same directory(sonarqube directory) as test files. As it was there already and never changed it didn't end up in the PR. :)

carminevassallo (approved):
LGTM!

Thanks for addressing the comments. I functionally validate the changes locally once again.

PR #900, opened 2026-07-24
SONAR-31124 Fix GCP Marketplace test

hatem-amairi-sonarsource:
## Problem
GCP Marketplace's `TEST_K8S_APP_FUNCTIONALITY` fails: the deployer Job exits with `BackoffLimitExceeded` on all GKE versions. `helm template` fails because `data-test/schema.yaml` defaults `jdbcOverwrite.jdbcUrl` to `jdbc:postgresql://myPostgres/myDatabase`, the placeholder that the chart's `validation.yaml` rejects.

## Changes
- `data-test/schema.yaml`: point the jdbcUrl default at a real value instead of the placeholder.
- `data-test/chart/templates/test-postgres.yaml` (new): ephemeral Postgres. The deployer applies it only through the `--mode=test` overlay. This is needed as GCP perform's a self-contained test and cannot connect to an external DB (a DB is required to run DCE).
- `Dockerfile`: package the overlay into `/data-test/chart/chart.tar.gz`.
- `TestMarketplaceDataTestSchemaRenders`: renders the chart with the real `data-test` defaults; fails if the placeholder returns.
- New CI job `verify-gcp-staging-app-selfcontained`: runs `mpdev verify` with no external DB and no param override, matching GCP path.

## Testing
- Unit test and local `helm template` pass.
- kind e2e with the overlay Postgres.
- Full GCP Marketplace CI on GKE, both verify GCP jobs green: https://github.com/SonarSource/helm-chart-sonarqube/actions/runs/30108210930

----
## Summary by Gitar

- **Tests:**
    - Added unit test asserting the GCP Marketplace data-test schema successfully renders the Helm chart
- **GCP Marketplace:**
    - Fixed functionality test by updating default JDBC URL to a valid endpoint and adding an ephemeral test-only PostgreSQL overlay

<sub>This will update automatically on new commits.</sub>

### DIFF
--- a/.github/workflows/gcp-marketplace.yml
+++ b/.github/workflows/gcp-marketplace.yml
@@ -128,8 +128,10 @@ jobs:
   # Mirrors Google's functionality test: no external database and no parameter override, so the
   # deployer relies on the data-test schema defaults and the /data-test/chart PostgreSQL overlay.
   # mpdev verify provisions and tears down its own ephemeral apptest-* namespace.
+  # Runs after (not alongside) verify-gcp-staging-app so only one DCE stack occupies the
+  # autoscaling staging cluster at a time, avoiding node-cap contention.
   verify-gcp-staging-app-selfcontained:
-    needs: [build-gcp-staging-app]
+    needs: [verify-gcp-staging-app]
     runs-on: github-ubuntu-latest-s
     name: Verify GCP Staging App (self-contained)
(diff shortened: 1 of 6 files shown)


carminevassallo (approved):
This change looks great and it solves a gap we had in the past nicely.

I'm approving this, let's hope it will solve the validation issue.

PR #909, opened 2026-08-10
SONAR-31416 Wire AGENTIC_ANALYSIS_URL so SonarQube polls Agentic Analysis health

lukasz-jarocki-sonarsource:
## Summary

Part of [EA-653](https://sonarsource.atlassian.net/browse/EA-653): SonarQube Server needs to know where to poll Agentic Analysis for health, the same way it already knows the Agent Orchestrator's URL.

- New `agenticHarness.agenticAnalysis.url` values.yaml key, independent of `agenticHarness.enabled` — Agentic Analysis is deployed and versioned separately from this chart (no Deployment/Service rendered here, unlike the orchestrator/runtimes), so this is a plain user-supplied URL rather than a computed in-cluster service name.
- Wires it into the SonarQube app container as `SONAR_AGENTICANALYSIS_URL` (maps to `sonar.agenticanalysis.url`), only rendered when the value is set.
- Documented in the chart README alongside the existing orchestrator-URL docs.

Targets `feature/unified-agentic-harness` (not `master`), matching where the rest of the agentic harness work lives — `master` doesn't have any agentic templates yet.

Companion changes in `sonar-enterprise` (new core-extension that polls this URL) and `sonarqube-unification` (Agent Orchestrator's own health/agents_status endpoints).

## Test plan

- [x] `helm template` verified in three scenarios: `agenticHarness.agenticAnalysis.url` blank → env var omitted; set with `agenticHarness.enabled=false` → renders standalone; set together with the full harness enabled → all three env vars (`SONAR_HUNTERAGENT_ORCHESTRATOR_URL`, `SONAR_REMEDIATIONAGENT_ORCHESTRATOR_URL`, `SONAR_AGENTICANALYSIS_URL`) render correctly side by side.

[EA-653]: https://sonarsource.atlassian.net/browse/EA-653?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lukasz-jarocki-sonarsource:
Closing this in favor of a follow-up PR. Since this was opened, @hatem-amairi-sonarsource's SONAR-31376 (commit d0827657) added a full chart-managed Vortex Analysis deployment (`vortexAnalysis.enabled` → `SONAR_VORTEX_ANALYSIS_URL` computed from the in-cluster service), which supersedes this PR's premise that Agentic Analysis has no chart-managed Deployment here. Merging this as-is would risk a duplicate `SONAR_VORTEX_ANALYSIS_URL` env entry if both mechanisms were ever enabled together.

Separately, while investigating why none of the 4 agentic health-check sections (EA-653) report healthy on next.sonarqube.com despite the orchestrator being genuinely deployed there, I found that `SONAR_HUNTERAGENT_ORCHESTRATOR_URL`/`SONAR_REMEDIATIONAGENT_ORCHESTRATOR_URL`/`SONAR_VORTEX_ANALYSIS_URL` — all three, including Hatem's new one — are set as raw pod env vars only. SonarQube's env-var-override mechanism (`AppSettingsLoaderImpl` in sonar-enterprise) only honors a `SONAR_*` env var for a property that's either a built-in `ProcessProperties.Property` or already has a literal line in `conf/sonar.properties` — a bare custom property like these three is never even checked against the environment. That's exactly why Next's health checks report everything unconfigured even though the env vars are correctly set on the container.

Follow-up PR incoming that fixes this at the chart level (renders these into the ConfigMap-backed `conf/sonar.properties` too, not just pod env vars) and covers Hatem's Vortex wiring as well as the pre-existing orchestrator wiring.

PR #910, opened 2026-08-12
SONAR-31376 Add Vortex analysis to the DCE chart

hatem-amairi-sonarsource:
# SONAR-31376 Add Vortex analysis to the DCE chart

Adds Vortex analysis to `charts/sonarqube-dce` as a standalone service deployed alongside SonarQube. It is independent of the Agentic Harness.

**Disabled by default.**

## What it adds

- `vortexAnalysis` values block, Deployment, Service, Secret and NetworkPolicy templates
- Sets `sonar.vortex.analysis.url` on the application nodes via `SONAR_VORTEX_ANALYSIS_URL` when enabled
- Matching rules on the application nodes' NetworkPolicy, so they can reach the service and accept its calls back to the Web API
- `values.schema.json` entries, README documentation and Terratest coverage

## Usage

```yaml
vortexAnalysis:
  enabled: true
  image:
    repository: <repository>
    tag: <tag>
  sonarqubeToken:
    existingSecret: my-vortex-token
```

## Notes

- No default `resources`: recommended values are not defined yet.
- The Deployment uses the `Recreate` strategy, so an upgrade never runs two pods at once (as the pod is heavy on resources) .
- The pod can take several minutes to become ready on a first start, which the startup probe allows for.
- Only for DCE chart. Replicating this on `charts/sonarqube` is tracked separately in [SONAR-31375](https://sonarsource.atlassian.net/browse/SONAR-31375).

Fixtures are regenerated in a separate [commit](https://github.com/SonarSource/helm-chart-sonarqube/commit/7620cd9891218953283e76cbb028172d444e60de).

Target branch: `feature/unified-agentic-harness`.

[SONAR-31375]: https://sonarsource.atlassian.net/browse/SONAR-31375?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


charts/sonarqube-dce/templates/vortex-analysis.yaml
```diff
@@ -0,0 +1,105 @@
...
+      nodeSelector:
+{{ toYaml .Values.nodeSelector | indent 8 }}
+      {{- end }}
+      {{- if .Values.tolerations }}
+      tolerations:
+{{ toYaml .Values.tolerations | indent 8 }}
+      {{- end }}
+      {{- if .Values.affinity }}
+      affinity:
+{{ toYaml .Values.affinity | indent 8 }}
```
navyatejaotturu-sonarsource, line 102: If root level nodeSelectors are used, doesn't it mean everything goes on to the same nodes? In a way it is same as having no `nodeSelector` assuming the root level configuration is applied to every resource. Why can't vortex have it's own `nodeSelector/affinity` declarations in `values.yaml` given that the description says it's resource intensive, it would make sense to schedule this on a different node than app/search pods for example.

charts/sonarqube-dce/values.yaml
```diff
@@ -847,6 +847,73 @@ mcp:
...
+
+  # Update strategy for the Vortex analysis Deployment. Recreate avoids running two pods at once
+  # during an upgrade.
+  strategy:
+    type: Recreate
+
+  # Token Vortex analysis uses to call the SonarQube Web API. Required for analysis to succeed.
+  sonarqubeToken:
+    # The token itself. Ignored when existingSecret is set.
+    token: ""
```
navyatejaotturu-sonarsource, line 876: Providing a plain token, should we really allow this, assuming the values.yaml is stored in a public facing GitHub repo?

charts/sonarqube-dce/templates/vortex-analysis.yaml
```diff
@@ -0,0 +1,105 @@
...
+        app: {{ include "sonarqube.name" . }}-vortex-analysis
+        release: {{ .Release.Name }}
+    spec:
+      automountServiceAccountToken: {{ .Values.serviceAccount.automountToken }}
+      {{- if or .Values.ApplicationNodes.image.pullSecrets .Values.ApplicationNodes.image.pullSecret $vortex.image.pullSecrets }}
+      imagePullSecrets:
+        {{- if .Values.ApplicationNodes.image.pullSecret }}
+        - name: {{ .Values.ApplicationNodes.image.pullSecret }}
+        {{- end }}
+        {{- with .Values.ApplicationNodes.image.pullSecrets }}
```
navyatejaotturu-sonarsource: When there is a specific property for providing imagePullSecrets for vortex, `vortex.image.pullSecrets` , why is application node imagePullSecrets are also attached to vortex deployment?

navyatejaotturu-sonarsource (changes requested):
On a whole nice work! I have few comments that might require clarifications rather than changes.

charts/sonarqube-dce/values.yaml
hatem-amairi-sonarsource: Inline credentials are the norm here, not the exception: `searchNodes.searchAuthentication.keyStorePassword`, `userPassword`, `jdbcOverwrite.jdbcPassword`, and especially `agenticHarness.orchestrator.sonarqubeToken.token` all accept a literal. The goal here is to enable users to use inline secrets for testing etc... but also offers a k8s Secret alternative for production.

navyatejaotturu-sonarsource: In that case, can we explicitly mention this is for testing purposes and should not be used for production setup?

hatem-amairi-sonarsource: Will discuss with the team and a ticket will follow as this is a convention we have repository wise.

charts/sonarqube-dce/values.yaml
```diff
@@ -847,6 +847,83 @@ mcp:
...
+# sonar.vortex.analysis.url property.
+vortexAnalysis:
+  enabled: false
+
+  image:
+    repository: ""
+    tag: ""
+    pullPolicy: IfNotPresent
+    # If using a private repository, the imagePullSecrets to use. Only these are applied: the
+    # application nodes' imagePullSecrets are not reused for Vortex analysis.
```
navyatejaotturu-sonarsource: I think we can remove referencing application node `imagePullSecrets` here. I think it is altogether a separate section not related to vortex.

navyatejaotturu-sonarsource (approved):
Looks good to me! 👍 Approving with a minor comment

PR #911, opened 2026-08-13
SONAR-31416 Also emit agentic health-check URLs into conf/sonar.properties

lukasz-jarocki-sonarsource:
## Summary

Follow-up to #909 (closed — superseded by @hatem-amairi-sonarsource's SONAR-31376 chart-managed Vortex Analysis deployment). Found this bug while investigating why none of EA-653's 4 agentic health-check sections report healthy on next.sonarqube.com, despite the Agent Orchestrator being genuinely deployed there.

**Root cause:** `SONAR_HUNTERAGENT_ORCHESTRATOR_URL`, `SONAR_REMEDIATIONAGENT_ORCHESTRATOR_URL`, and `SONAR_VORTEX_ANALYSIS_URL` are all set as raw pod env vars only. SonarQube's `Configuration` API (`org.sonar.api.config.Configuration`) only honours a `SONAR_*` env var override for a property that is either a built-in `ProcessProperties.Property` or already has a literal line in `conf/sonar.properties` (`AppSettingsLoaderImpl.load()`/`loadPropertiesFromEnvironment()` in sonar-enterprise, `server/sonar-main`). A bare custom property with neither — like these three — is never even checked against the environment. This runs once in the launcher process before the WebServer JVM (and any CoreExtension) starts, so nothing later can fix it.

Confirmed on Next: Datadog logs show `Agent Orchestrator URL is not configured` / `Agentic Analysis URL is not configured` warnings on both DCE app pods, even though `agenticHarness.enabled: true` and the orchestrator is genuinely running in the same namespace, and the chart correctly sets the env vars.

This exact gotcha is already known and documented in `sonarqube-unification`'s `HunterAgentWebConfiguration.java`, which works around it by reading the property via Spring `@Value` (bypassing `Configuration` entirely) instead — but that's only viable for Spring-based modules. Plain SonarQube `Configuration` consumers (like `core-extension-agents`, EA-653's new health-check code) have no such escape hatch.

## Fix

Also emit `sonar.hunteragent.orchestrator.url`, `sonar.remediationagent.orchestrator.url`, and `sonar.vortex.analysis.url` into the ConfigMap-backed `conf/sonar.properties` (via a new `sonarqube.agenticHealthProperties` helper, merged into the existing `sonarqube.mergedSonarProperties` alongside the Hazelcast auto-properties), in addition to the existing pod env vars — additive, not a replacement, since `hunter-agent-unified-app` still needs the raw env var for its Spring-based read.

Verified via `helm template` with `agenticHarness.enabled`/`vortexAnalysis.enabled`:
- Both properties render correctly into `conf/sonar.properties` when enabled, and are absent when disabled.
- A user-supplied `ApplicationNodes.sonarProperties` entry for the same key still wins (existing override behavior preserved).
- No duplicate env vars introduced.

## Test plan

- [x] `helm template` with `agenticHarness.enabled=true` + `vortexAnalysis.enabled=true`: all 3 properties appear as `conf/sonar.properties` lines with the correct in-cluster URLs, and the existing pod env vars are unchanged (no duplicates).
- [x] `helm template` with both disabled: no agentic lines rendered.
- [x] `helm template` with a user-supplied `ApplicationNodes.sonarProperties["sonar.hunteragent.orchestrator.url"]` override: user value wins over the auto-generated one.
- [x] Existing Go chart tests (`tests/unit-test/`) — couldn't run locally (no Go toolchain available), relying on CI.

### DIFF
--- a/charts/sonarqube-dce/templates/sonarqube-application.yaml
+++ b/charts/sonarqube-dce/templates/sonarqube-application.yaml
@@ -100,7 +100,7 @@ spec:
             {{- . | toYaml | trim | nindent 12 }}
           {{- end }}
       {{- end }}
-      {{- $hasMergedProps := or .Values.ApplicationNodes.sonarProperties .Values.ApplicationNodes.port .Values.ApplicationNodes.webPort .Values.ApplicationNodes.cePort .Values.agenticHarness.enabled }}
+      {{- $hasMergedProps := or .Values.ApplicationNodes.sonarProperties .Values.ApplicationNodes.port .Values.ApplicationNodes.webPort .Values.ApplicationNodes.cePort .Values.agenticHarness.enabled .Values.vortexAnalysis.enabled }}
       {{- if or $hasMergedProps .Values.ApplicationNodes.sonarSecretProperties .Values.sonarSecretKey }}
         - name: concat-properties
           image: {{ default (include "sonarqube.image" .) .Values.initContainers.image }}
@@ -325,6 +325,7 @@ spec:
               value: {{ .Values.mcp.healthCheckInterval | quote }}
             {{- end }}
             {{- end }}
+            {{- /* Also given a conf/sonar.properties line via mergedSonarProperties (SONAR-31416). */}}
             {{- if .Values.vortexAnalysis.enabled }}
             - name: SONAR_VORTEX_ANALYSIS_URL
               value: {{ include "sonarqube.vortexAnalysis.url" . | quote }}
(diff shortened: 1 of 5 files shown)


charts/sonarqube-dce/templates/sonarqube-application.yaml
hatem-amairi-sonarsource: Can we also add `.Values.vortexAnalysis.enabled` here
It is needed so the ConfigMap is mounted

lukasz-jarocki-sonarsource: Good catch — fixed. `$hasMergedProps` only checked `agenticHarness.enabled`, so a vortexAnalysis-only deployment (no agenticHarness) would write `sonar.vortex.analysis.url` into the ConfigMap but never mount `conf/sonar.properties` at all, silently defeating the property. Added `.Values.vortexAnalysis.enabled` to the same `or`. Verified with `helm template` on the vortex-analysis-only fixture: the concat-properties init container, the config/concat-dir volumes, and the conf/sonar.properties mount now all render as expected.

hatem-amairi-sonarsource (approved):
LGTM!

PR #916, opened 2026-08-18
SONAR-31523,SONAR-31525 Fix orchestrator init container pull secrets; document runtime storage egress

hatem-amairi-sonarsource:
## Summary
- SONAR-31523: the orchestrator's `wait-for-sonarqube` init container pulls the SonarQube
  image but only ever got the orchestrator's own pull secrets, never
  `applicationNodes.image.pullSecret(s)` — unlike the identical init container in `mcp.yaml`.
  Fixed to include both.
- SONAR-31525: job runtimes (`hunter`/`remediation`) read/write artifacts directly against
  object storage via presigned URLs, bypassing the orchestrator, but nothing documented that
  their NetworkPolicy `egressAllow` must cover it. Documented in `values.yaml` and the README.

## Test plan
- Added/extended Go unit tests covering pull-secret combinations and egress rendering for
  both runtime families (`go test ./...` passing).
- Verified both changes against a live NetworkPolicy-enforcing test cluster, with no
  regressions.

### DIFF
--- a/charts/sonarqube-dce/templates/agentic-orchestrator.yaml
+++ b/charts/sonarqube-dce/templates/agentic-orchestrator.yaml
@@ -1,3 +1,5 @@
+{{ $applicationNodes := fromYaml (include "applicationNodes" . ) }}
+{{ $_ := set .Values "ApplicationNodes" $applicationNodes }}
 {{- if .Values.agenticHarness.enabled }}
 apiVersion: apps/v1
 kind: Deployment
--- a/charts/sonarqube-dce/templates/agentic-orchestrator.yaml
+++ b/charts/sonarqube-dce/templates/agentic-orchestrator.yaml
@@ -37,7 +37,7 @@ spec:
         - name: {{ .Values.agenticHarness.orchestrator.image.pullSecret }}
         {{- end }}
         {{- if .Values.agenticHarness.orchestrator.image.pullSecrets }}
-{{ toYaml .Values.agenticHarness.orchestrator.image.pullSecrets | indent 8 }}
+        {{- toYaml .Values.agenticHarness.orchestrator.image.pullSecrets | nindent 8 }}
         {{- end }}
       {{- end }}
       {{- with .Values.agenticHarness.orchestrator.securityContext }}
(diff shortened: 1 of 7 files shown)


charts/sonarqube-dce/templates/agentic-orchestrator.yaml
```diff
@@ -23,8 +23,16 @@ spec:
         sonarqube.agentic/component: orchestrator
     spec:
       automountServiceAccountToken: {{ if .Values.agenticHarness.serviceAccount.create }}{{ .Values.agenticHarness.serviceAccount.automountToken }}{{ else }}{{ .Values.serviceAccount.automountToken }}{{ end }}
-      {{- if or .Values.agenticHarness.orchestrator.image.pullSecrets .Values.agenticHarness.orchestrator.image.pullSecret }}
+      {{- if or .Values.ApplicationNodes.image.pullSecrets .Values.ApplicationNodes.image.pullSecret .Values.agenticHarness.orchestrator.image.pullSecrets .Values.agenticHarness.orchestrator.image.pullSecret }}
```
navyatejaotturu-sonarsource, line 28: Isn't `ApplicationNodes` with capital `A` deprecated in the `values.yaml`? I think we should adapt this. Same applies at other places where this is used.

navyatejaotturu-sonarsource (changes requested):

charts/sonarqube-dce/templates/agentic-orchestrator.yaml
hatem-amairi-sonarsource: `ApplicationNodes` here is intentional — it's the chart's deep-merge compat pattern, not raw use of the deprecated key. `_helpers.tpl`'s `applicationNodes` template merges `.Values.applicationNodes` with the deprecated `.Values.ApplicationNodes`, and several templates (http-route.yaml, ingress.yaml, sonarqube-search.yaml, etc.) run that merge unconditionally and write it back onto `.Values.ApplicationNodes` for the whole chart render. Other templates (mcp.yaml, sonarqube-application.yaml, application.yaml, change-admin-password-hook.yaml) then just read `.Values.ApplicationNodes.image.*` directly — this file follows that same existing pattern. Switching to lowercase here would actually drop support for the deprecated key that every other template still honors.

navyatejaotturu-sonarsource (approved):

PR #918, opened 2026-08-19
SONAR-31686,SONAR-31689 Default gVisor sandboxing to opt-out; Remove agenticHarness.*

hatem-amairi-sonarsource:
## SONAR-31686:  gVisor opt-out

- gVisor sandboxing for the agentic runtimes flips from opt-in to opt-out: `gvisor.enabled` / `gvisor.installer.enabled` now default to `true` on the DCE chart (explicit overrides still win; see SONAR-31689 below for the exact gating condition after the `agenticHarness.*` rename).
- gVisor is dropped from the standard `sonarqube` chart entirely — its agentic runtimes aren't ported there yet, so shipping an unused RuntimeClass/installer was unnecessary scope.
- README corrected: opt-out framing, isolation trade-off, and a narrowed managed-EKS/AKS installer claim.
- Unit tests cover all toggle combinations; compatibility fixtures regenerated and kubeconform-clean.
- Live-verified all combinations on a local test cluster.

## SONAR-31689:  removal of agenticHarness.*

- Replace the `agenticHarness` umbrella values key with four independent top-level keys —
  `orchestrator`, `hunterAgent`, `remediationAgent`, `gvisor` — each with its own `.enabled`,
  sibling to the existing `vortexAnalysis`.
- Add fail-fast dependency validation: `hunterAgent.enabled` requires `orchestrator.enabled`;
  `remediationAgent.enabled` requires `orchestrator.enabled` AND `vortexAnalysis.enabled`.
- gVisor now renders only when `gvisor.enabled` AND at least one of `hunterAgent`/`remediationAgent`
  is enabled (previously gated on the removed `agenticHarness.enabled`).
- `serviceAccount`/`networkPolicy` config move from shared, cross-component toggles to
  self-contained per-component blocks, matching how `vortexAnalysis` already works.
- Scope the CI "Verify gVisor installed" step to the `sonarqube-dce` chart only (the standard
  `sonarqube` chart no longer has gVisor); update `ci-values.yaml` to enable `orchestrator` +
  `hunterAgent` (replicaCount 0, images aren't pullable from CI) so gVisor's new dependency-based
  gating still renders and gets exercised in CI.
- Clean rename, no back-compat aliasing.

## Test plan
- [x] `helm lint charts/sonarqube-dce`
- [x] `go test ./unit-test/...`, incl. new dependency-validation and gVisor render-matrix coverage
- [x] Compatibility fixtures regenerated
- [x] Deployed end-to-end on a real cluster: orchestrator, hunter, remediation, gVisor and
      vortex-analysis all healthy


charts/sonarqube-dce/ci/ci-values.yaml
```diff
@@ -41,7 +41,25 @@ tests:
       memory: 200M
       ephemeral-storage: 1000M
 
+# gVisor only renders when the harness is enabled. The orchestrator and runtime images are not
+# pullable from CI, so both run with replicaCount 0: the harness is on here purely to exercise the
+# gVisor RuntimeClass and its installer DaemonSet on the kind worker.
```
navyatejaotturu-sonarsource: Why do we need to enable gvisor if agentic pods are not deployed here?

charts/sonarqube-dce/ci/ci-values.yaml
```diff
@@ -41,11 +41,29 @@ tests:
...
+    tag: "0.0.0"
+  storage:
+    bucket: "ci"
+
+hunterAgent:
+  enabled: true
+  replicaCount: 0
+  image:
+    repository: "example.com/hunteragent"
+    tag: "0.0.0"
```
navyatejaotturu-sonarsource: Do we really need this? I was assuming that gvisor is by default enabled without these?

navyatejaotturu-sonarsource (approved):
Looks good to me!

PR #919, opened 2026-08-20
SONAR-31689 Remove agenticHarness.* from the Helm charts

hatem-amairi-sonarsource:
## Summary
- Replace the `agenticHarness` umbrella values key with four independent top-level keys —
  `orchestrator`, `hunterAgent`, `remediationAgent`, `gvisor` — each with its own `.enabled`,
  sibling to the existing `vortexAnalysis`.
- Add fail-fast dependency validation: `hunterAgent.enabled` requires `orchestrator.enabled`;
  `remediationAgent.enabled` requires `orchestrator.enabled` AND `vortexAnalysis.enabled`.
- gVisor now renders only when `gvisor.enabled` AND at least one of `hunterAgent`/`remediationAgent`
  is enabled (previously gated on the removed `agenticHarness.enabled`).
- `serviceAccount`/`networkPolicy` config move from shared, cross-component toggles to
  self-contained per-component blocks, matching how `vortexAnalysis` already works.
- Drop the CI "Verify gVisor installed" step: unreachable now that `ci-values.yaml` enables no
  agent, so gVisor never renders there under the new gating.
- Clean rename, no back-compat aliasing.

## Test plan
- [x] `helm lint charts/sonarqube-dce`
- [x] `go test ./unit-test/...`, incl. new dependency-validation and gVisor render-matrix coverage
- [x] Compatibility fixtures regenerated
- [x] Deployed end-to-end on a real cluster: orchestrator, hunter, remediation, gVisor and
      vortex-analysis all healthy


charts/sonarqube-dce/templates/agentic-serviceaccount.yaml
```diff
@@ -37,12 +37,10 @@ metadata:
...
+  {{- with $cfg.serviceAccount.annotations }}
   annotations:
-{{ toYaml $rann | indent 4 }}
+{{ toYaml . | indent 4 }}
   {{- end }}
-automountServiceAccountToken: {{ $.Values.agenticHarness.serviceAccount.automountToken }}
-{{- end }}
+automountServiceAccountToken: {{ $cfg.serviceAccount.automountToken }}
 {{- end }}
 {{- end }}
```
navyatejaotturu-sonarsource, line 46: Just a refactoring suggestion, why can't everything go in to the `dict` and then rendered. Why `orchestrator`'s service account is rendered separately

navyatejaotturu-sonarsource (approved):
Looks good to me! However, I have a refactoring suggestion below.

charts/sonarqube-dce/templates/agentic-serviceaccount.yaml
hatem-amairi-sonarsource: This will be done as part of [SONAR-31655](https://sonarsource.atlassian.net/browse/SONAR-31655)

PR #923, opened 2026-08-21
SONAR-31721 Pre-flight check for gVisor installer

hatem-amairi-sonarsource:
## Summary
- The gVisor installer DaemonSet previously always attempted to write runsc's
  containerd config, even on nodes that already have a sandbox runtime
  registered under the same handler (e.g. GKE Sandbox) — risking a duplicate/
  conflicting containerd config block and a needlessly failing install.
- Add a pre-flight check to install-gvisor.sh: if crictl reports the target
  handler already live, skip the whole install step before touching disk.
  The RuntimeClass and node-labeling step are unaffected, so scheduling still
  works normally.

## Test plan
- [x] `go test ./...` in `tests/unit-test` (new `TestGvisorInstallerPreflightSkipsExistingRuntime` + full suite pass)
- [x] Rendered the chart standalone and confirmed the check lands in the ConfigMap before the download step
- [x] Ran the rendered script locally under a mocked `crictl`: skips cleanly when the handler is present, falls through unchanged when it isn't
- [x] Deployed to a dev EKS cluster (fresh install): all pods healthy, gvisor-installer DaemonSet running
- [x] On a live node, temporarily faked `crictl` to report the handler present, re-ran the real `install-gvisor.sh`, confirmed it skips immediately; removed the fake binary and verified node state (runsc binary, containerd config) untouched


carminevassallo (approved):
LGTM!

PR #924, opened 2026-08-25
SONAR-31375 Port the agentic pack to the sonarqube chart

carminevassallo:
## Summary

Ports the Agent Orchestrator, hunter/remediation runtimes, Vortex analysis, and gVisor sandboxing from `charts/sonarqube-dce` to `charts/sonarqube`, at full parity. Every agentic feature defaults to `enabled: false`, except `gvisor.enabled: true` (inert on its own — `sonarqube.gvisor.enabled` also requires a runtime to be enabled).

- `charts/sonarqube-dce/` is untouched except for one comment and one README sentence: the gVisor `RuntimeClass` has a fixed default name and is cluster-scoped, so now that both charts ship the feature, the existing "several releases of this chart collide" warning had to mention the cross-chart case too. No rendered output changes there. It remains the reference implementation.
- With every agentic key at its default, `helm template charts/sonarqube` is byte-identical to the base branch.
- 33 pre-existing compatibility fixtures under `tests/unit-compatibility-test/fixtures/sonarqube/` are unchanged; two new ones added for the agentic pack.

See `charts/sonarqube/README.md`'s new `### Agents` section for user-facing docs.

### Two deliberate divergences from DCE

1. **NetworkPolicy egress.** This chart already allows unrestricted egress, so no orchestrator/Vortex-specific egress rules were added — only the matching ingress rules. DCE needs both.
2. **`agentOrchestrator.enabled` requires `jdbcOverwrite.enabled=true`.** The orchestrator reads and writes SonarQube's own database, and the embedded database this chart selects by default is not reachable from another pod. `templates/validation.yaml` fails the install rather than let it deploy against a database that cannot work. DCE has no such gate — it is always on an external database. Every `CORE_DB_*` value still defaults to the matching `jdbcOverwrite` setting and can be overridden individually via `agentOrchestrator.coreDb.*`; those overrides supplement `jdbcOverwrite`, they do not replace it.

Because of (2), the agentic pack is **not** exercised by this chart's `ct install` job — that job installs with the embedded database (`community.enabled`) and is worth keeping that way. gVisor only renders when a runtime is enabled, and a runtime requires the orchestrator, so there is no way to cover it here without also standing up PostgreSQL. `charts/sonarqube/ci/ci-values.yaml` is therefore unchanged by this PR, and the DCE kind job remains the end-to-end gVisor verification. The templates are byte-identical between the two charts, so the coverage lost is marginal.

## Test plan

- [x] `ct lint --config ct-sonarqube-test.yaml` and `ct lint --config ct-sonarqube-dce-test.yaml`
- [x] `helm template` byte-identical to the base branch at agentic defaults
- [x] Go unit tests (`tests/unit-test`) pass, including new coverage for the `jdbcOverwrite` requirement and for the agent URLs landing in the `sonar.properties` ConfigMap (the SONAR-31416 fix — previously only the pod env var was asserted)
- [x] Compatibility fixtures regenerated with zero drift on the 33 existing fixtures
- [x] `values.schema.json` validates and renders
- [x] `diff` of all 12 agentic templates against their DCE counterparts is empty, apart from DCE's `ApplicationNodes` preamble in `agent-orchestrator.yaml`

Not run: bare `helm lint charts/sonarqube` fails at chart defaults on a pre-existing `sonarqube-sts.yaml` parse error, identically on the base branch. `ct lint`, which lints against `ci/ci-values.yaml`, is the gate CI actually runs.


charts/sonarqube-dce/README.md
hatem-amairi-sonarsource, line 1106: (Unrelated to the PR): Does SQ send requests to Vortex ? (Isn't only the other way around: Vortex -> SQ)

hatem-amairi-sonarsource, line 969: (Unrelated to the PR): Does Vortex belong under `### Agents` section ?

charts/sonarqube/README.md
```diff
@@ -862,6 +862,161 @@ and set `persistence.hostPath.path` and `persistence.hostPath.type`.
 | `mcp.resources`                        | CPU/memory resource requests and limits for the MCP container                                            | `{}`                                                                   |
 | `mcp.annotations`                      | Annotations for the MCP pod                                                                              | `{}`                                                                   |
 
+### Agents
+
+Four independently toggled features: Vortex (`vortex.enabled`), the shared Agent Orchestrator (`agentOrchestrator.enabled`), and one Deployment/Service each for the Hunter Agent (`hunterAgent.enabled`) and the Remediation Agent (`remediationAgent.enabled`). Dependencies are enforced at template time (see `templates/validation.yaml`): `hunterAgent.enabled=true` requires `agentOrchestrator.enabled=true`; `remediationAgent.enabled=true` requires both `agentOrchestrator.enabled=true` and `vortex.enabled=true`.
```
hatem-amairi-sonarsource, line 867: (Same as above for the dce-chart, unrelated to the PR).

charts/sonarqube/README.md
```diff
@@ -862,6 +862,161 @@ and set `persistence.hostPath.path` and `persistence.hostPath.type`.
...
+| `hunterAgent.scriptPath`                                         | Detection-agent entrypoint exposed as `SCRIPT_PATH`, which switches the runtime into hunter/detection mode; `""` omits the env var                        | `/home/agent/app/.venv/bin/detection-agent`                                    |
+| `remediationAgent.remediationScriptPath`                         | Remediation-agent entrypoint exposed as `REMEDIATION_SCRIPT_PATH` (`REMEDIATION_RULE_INFO_ENDPOINT` is derived from the orchestrator URL); `""` omits the env var | `/home/agent/app/.venv/lib/python3.13/site-packages/remediation_agent/main.py` |
+| `remediationAgent.resources`                                     | Resources for the SonarQube Remediation Agent (SQRA) runtime, sized for repo clone + agent/LLM loop                                      | requests `1` CPU / `2Gi` / `10Gi`, limits `4` CPU / `8Gi` / `50Gi`             |
+
+When `vortex.enabled` is set to `true`, the chart deploys a separate Vortex pod alongside the SonarQube application nodes, and SonarQube starts sending analysis requests to it. While it is disabled, SonarQube sends none.
```
hatem-amairi-sonarsource, line 1008: (Same as above for the dce-chart, unrelated to the PR).

hatem-amairi-sonarsource (approved):
LGTM 👍 
Tested on my side also on EKS.

charts/sonarqube-dce/README.md
carminevassallo: Thanks for the comments. We'll review our documentation before merging to master. Noted down your points.

PR #926, opened 2026-08-27
SONAR-31746 Mount the encryption key to the orchestrator on the sonarqube chart

carminevassallo:
## Summary

Ports the DCE chart's encryption-key mount for the agent orchestrator
(originally implemented only for `charts/sonarqube-dce`) to the standard
`charts/sonarqube` chart.

- `charts/sonarqube/templates/agent-orchestrator.yaml`: when
  `.Values.sonarSecretKey` is set, mount it into the orchestrator and expose
  its path via `AGENTIC_SECRET_KEY_PATH` (`/sonarcloud/secret/sonar-secret.txt`),
  mirroring the main SonarQube pod's `sonar.secretKeyPath` mount.
- `charts/sonarqube/values.yaml`: document the new behavior next to
  `sonarSecretKey`.
- `tests/unit-test/agent_orchestrator_test.go`: fix
  `TestAgentOrchestratorEncryptionKeyMount`, left broken (wrong call arity to
  `renderAgentOrchestrator`) since the `SONAR-31375` chart port, so it covers
  both charts via the existing `agentCharts` parametrization.

## Test plan

- [x] `go vet ./...`
- [x] `go test -run TestAgentOrchestrator -v ./...` (both `sonarqube-dce` and
      `sonarqube` subtests pass)
- [x] `go test ./...` (full suite passes)
- [x] Diffed `helm template` output against committed golden fixtures for the
      affected test-case values files (`remediation-agent-values.yaml`,
      `config-values.yaml`) — no change, no fixture regeneration needed

Jira: https://sonarsource.atlassian.net/browse/SONAR-31746


matteo-mara-sonarsource (approved):
LGTM!

PR #928, opened 2026-08-28
SONAR-31879 Default fsGroup and HOME for MCP pod so /data is writable by non-root MCP process

jimil09:
----
## Summary by Gitar

- **MCP Configuration:**
    - Configured default `fsGroup: 0` in `podSecurityContext` and set `HOME=/data` environment variable for the MCP pod
    - Added support for `mcp.initContainers` to handle storage drivers that ignore `fsGroup` restrictions

<sub>This will update automatically on new commits.</sub>


navyatejaotturu-sonarsource (approved):
Looks good to me!

PR #938, opened 2026-09-02
SONAR-32065 Derive and mount agentic signing keys

carminevassallo:
Jira: [SONAR-32065](https://sonarsource.atlassian.net/browse/SONAR-32065)

One operator-supplied instance secret is expanded by a pre-install hook Job into one HKDF-derived key per communication hop, and each pod mounts only the hops it takes part in. A component holding a key for a hop it isn't on is the failure mode this feature exists to prevent, so the distribution matrix is pinned by tests rather than left to review.

| Label | Hop | Derived when | Mounted on |
|---|---|---|---|
| `orchestrator-to-hunter` | Orchestrator ↔ Hunter | `hunterAgent.enabled` | Orchestrator, Hunter |
| `orchestrator-to-remediation` | Orchestrator ↔ Remediation | `remediationAgent.enabled` | Orchestrator, Remediation |
| `remediation-to-sqs` | Remediation → SQS | `remediationAgent.enabled` | Remediation only |
| `agentic-shared` | SQS ↔ Orchestrator, Vortex → SQS | any agentic component | SQS, Orchestrator, Vortex |

## What lands in the pods

Agent-side components share one neutral directory since they are three different base images; SQS keeps its material under `sonarqubeFolder` alongside the existing `secret/` mount. Every variable points at an individual file, never a directory, and is only emitted when the file behind it is actually projected into that pod — except on SQS, whose `agentic-shared` file is read through a property instead of a variable at all; see below.

| Consumer | Mount dir | Env vars |
|---|---|---|
| Orchestrator | `/etc/agentic/keys` | `AGENTIC_HUNTER_RUNTIME_SIGNING_KEY_PATH`, `AGENTIC_REMEDIATION_RUNTIME_SIGNING_KEY_PATH`, `AGENTIC_SONARQUBE_SIGNING_KEY_PATH` |
| Hunter | `/etc/agentic/keys` | `AGENTIC_VERIFY_KEY_PATH`, `AGENTIC_VERIFY_KEY_ID=orchestrator-to-hunter` |
| Remediation | `/etc/agentic/keys` | `AGENTIC_VERIFY_KEY_PATH`, `AGENTIC_VERIFY_KEY_ID=orchestrator-to-remediation`, `REMEDIATION_AGENTIC_SIGNING_KEY_PATH` |
| SQS | `/opt/sonarqube/agentic-keys` (keys), `/opt/sonarqube/agentic-secret` (instance secret) | `AGENTIC_SIGNING_SECRET_FILE` |
| Vortex | `/etc/agentic/keys` | `AGENTIC_ORCHESTRATOR_SIGNING_KEY_PATH` |

`AGENTIC_VERIFY_KEY_PATH` is the same name on both runtimes and doesn't collide — each pod mounts exactly one verification key. Because the name carries no hop, it is paired with `AGENTIC_VERIFY_KEY_ID`, whose value is the label itself: that is what tells the runtime which hop it is verifying.

The orchestrator mounts `agentic-shared` and now points at it with `AGENTIC_SONARQUBE_SIGNING_KEY_PATH`, so orchestrator→SQS calls get signed. SQS mounts the same key but gets no variable for it at all — see "SQS is the asymmetric consumer" below for why.

## The hook Job

Runs `derive-keys.sh`, which ships inside the Agent Orchestrator image so air-gapped installs need no second image pull. `agentKeyDerivation.image` overrides that for a Vortex-only install, which needs derived keys but has no orchestrator to borrow the script from — without it the release would have to pull an orchestrator it never deploys.

- `pre-install,pre-upgrade`, weight `0`, `hook-succeeded` deletion, bounded by `activeDeadlineSeconds` so it can't hang a release
- Istio injection disabled — a sidecar would keep the pod alive and the hook would never succeed
- Writes the per-consumer Secrets through the Kubernetes API under a namespaced Role limited to `get`/`create`/`update` on `secrets` (never `list`/`watch`/`delete`, never a ClusterRole), using its own ServiceAccount
- The instance secret reaches it as a mounted volume, never an env var or argument
- Deterministic: nothing release-specific is passed in, so re-running on upgrade doesn't rotate every key and restart every agentic pod
- `agentKeyDerivation.enabled=false` is the escape hatch for operators who derive the keys themselves — no Job and no Role, but the consumers still mount the same Secret names
- `agentKeyDerivation.serviceAccount.create=false` runs the Job under `agentKeyDerivation.serviceAccount.name`, falling back to the release's top-level `serviceAccount`; the chart binds the Role to whichever it lands on. The override is read on that path precisely because it is what the fail-closed message tells operators to set — ignoring it would make following the message return the identical failure.

## SQS is the asymmetric consumer

It mounts the *instance secret* and derives `remediation-to-sqs` and `agentic-shared` in-process to **verify** inbound calls (`sonar.agentic.signing.secretFile`) — which is why it mounts neither of those keys. Separately it **signs** its own calls to the orchestrator with the mounted `agentic-shared` key (`sonar.agentic.orchestrator.signingKeyPath`, added by SONAR-31996; blank leaves those calls unsigned rather than failing).

Both are written into `conf/sonar.properties`, because a property with a properties-file line ignores its env override (SONAR-31416). `secretFile` keeps a redundant `AGENTIC_SIGNING_SECRET_FILE` env var carrying the same value, and a test cross-checks the two match. `signingKeyPath` has no such companion at all — SQS is the only consumer of a mounted key with no matching `AGENTIC_*` variable, since one would just go unread by `Configuration.get()`. Each property is instead resolved directly against what the pod actually projects, so nothing but that test stops the property and the mount drifting apart silently.

## Failing closed

Enabling anything agentic without `agenticSigningSecret.existingSecret` fails the render: there is nothing to derive from, so the components would start unable to sign or verify anything. Same for the hook having no image anywhere — keys that never get derived would surface as pods stuck on a missing Secret, or worse, on upgrade, as silently stale ones.

## Scope

Both charts: `charts/sonarqube` and `charts/sonarqube-dce`. The helpers, hook, RBAC and validation are identical; the only difference is where the mounts land — on DCE they belong on `sonarqube-application.yaml` alone, since agentic traffic only ever targets Application Nodes.

## CI

The DCE `ci-values.yaml` already enabled a runtime at `replicaCount: 0` purely so gVisor renders for the DCE-only "verify gVisor installed" step. That now also switches on the signing path, and the image carrying `derive-keys.sh` isn't pullable from CI — so `agentKeyDerivation.enabled=false` there and the workflow pre-creates the Secrets the Job would have written with throwaway values. That is the same "bring your own keys" mode the values document for GitOps and sealed-secrets users, so the install exercises it rather than skipping it. `fullnameOverride` pins the derived Secret names, which are otherwise built from the release name chart-testing generates at random.

`sonar-project.properties` excludes `tests/unit-compatibility-test/fixtures/**` — generated fixtures, counted as duplication by the quality gate.

## Verification

- `go test -C tests/unit-test ./...` — green, including `agent_key_derivation_test.go` covering the hook, the RBAC, the distribution matrix, the per-pod mounts, the ServiceAccount override, and a cross-hop invariant — plus new subtests asserting `signingKeyPath` resolves to a projected file with no companion env var on SQS, alongside the pre-existing `secretFile`/env-var cross-check
- `helm lint` clean on both charts against their `ci-values.yaml`; DCE renders 32 resources with no duplicate `(kind, name)` and nothing over 63 characters
- The four secret references in the DCE render match the four Secrets the workflow creates, exactly
- `./.github/scripts/generate_helm_fixtures.sh` produces no diff
- Negative-checked the property/mount cross-check by mutating the helper — it fails as intended
- Deployed both charts to a live cluster (speedboat-test-xx) with the current image; confirmed the orchestrator picks up `AGENTIC_SONARQUBE_SIGNING_KEY_PATH` and SQS no longer carries `AGENTIC_ORCHESTRATOR_SIGNING_KEY_PATH`, reading the key path from `conf/sonar.properties` instead

[SONAR-32065]: https://sonarsource.atlassian.net/browse/SONAR-32065?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


carminevassallo:
@matteo-mara-sonarsource I'm asking you for a review of the overall list of env vars (see the PR description). In addition to the table that was sent by Manuel, I was adding `AGENTIC_VERIFY_KEY_ID` (it was listed in the specification document, so it should be ok) and I mount the agentic-shared-key to both vortex and SQS. Could you please double check this part and tell me how to change the env var names in case?

charts/sonarqube-dce/templates/agent-key-derivation-rbac.yaml
```diff
@@ -0,0 +1,80 @@
...
+    sonarqube.agent/component: key-derivation
+    app.kubernetes.io/component: agent-key-derivation
+  annotations:
+    "helm.sh/hook": pre-install, pre-upgrade
+    "helm.sh/hook-weight": "-5"
+    "helm.sh/hook-delete-policy": before-hook-creation
+rules:
+  - apiGroups: [""]
+    resources: ["secrets"]
+    verbs: ["get", "create", "update"]
```
hatem-amairi-sonarsource, line 55: `get` does not seem to be needed for the derivation process.
Also we should scope this to just the keys (least permission), very important here as this role will survive a helm uninstall.

charts/sonarqube-dce/templates/_helpers.tpl
```diff
@@ -806,6 +815,280 @@ true
...
+*/}}
+{{- define "sonarqube.agentKeyDerivation.required" -}}
+{{- if and .Values.agentKeyDerivation.enabled (eq (include "sonarqube.agentic.enabled" .) "true") .Values.agenticSigningSecret .Values.agenticSigningSecret.existingSecret -}}
+true
+{{- end -}}
+{{- end -}}
+
+{{- define "sonarqube.agentKeyDerivation.fullname" -}}
+{{- printf "%s-agent-key-derivation" (include "sonarqube.fullname" .) | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
```
hatem-amairi-sonarsource, line 1052: This helper uses a plain trunc 63 so a 52–53 char release name the Job, Role, RoleBinding and ServiceAccount all render as `sonarqube.fullname` and collide

matteo-mara-sonarsource:
> @matteo-mara-sonarsource I'm asking you for a review of the overall list of env vars (see the PR description). In addition to the table that was sent by Manuel, I was adding `AGENTIC_VERIFY_KEY_ID` (it was listed in the specification document, so it should be ok) and I mount the agentic-shared-key to both vortex and SQS. Could you please double check this part and tell me how to change the env var names in case?

After checking this a few times I can say that the orchestrator is missing the AGENTIC-SHARED key according to the table in the description is getting access only to the remediation and the hunter agent keys. 
The master instance secret only on SQS.

charts/sonarqube-dce/templates/agent-key-derivation-hook.yaml
```diff
@@ -0,0 +1,173 @@
...
+              mountPath: /tmp
+      volumes:
+        - name: instance-secret
+          secret:
+            secretName: {{ .Values.agenticSigningSecret.existingSecret }}
+            items:
+              - key: {{ include "sonarqube.agentic.signingSecretKey" . }}
+                path: instance-secret
+        - name: tmp
+          emptyDir: {}
```
hatem-amairi-sonarsource, line 172: SQS, the orchestrator and the runtimes all accept a `.previous` sibling for rolling rotation, and `derive-keys.sh` emits one when `<secret-file>.previous` exists. This single-item `items:` projection means it never can, so no `.previous key` is ever derived — and `write_secret` (`line 122`), `keyVolume` (`_helpers.tpl:1023`) and the app-node instance-secret volume (`sonarqube-application.yaml:521-526`) would drop it anyway.

Rotating the instance secret is therefore a hard cutover: during the rollout, old- and new-key pods coexist and cross-component calls fail. Worth wiring, or at least documenting.

charts/sonarqube-dce/templates/agent-key-derivation-hook.yaml
```diff
@@ -0,0 +1,173 @@
...
+      automountServiceAccountToken: true
+      {{- $image := fromYaml (include "sonarqube.agentKeyDerivation.image" .) }}
+      {{- $pullSecrets := include "sonarqube.agent.imagePullSecrets" (list $image) }}
+      {{- if $pullSecrets }}
+      imagePullSecrets: {{- $pullSecrets | nindent 8 }}
+      {{- end }}
+      serviceAccountName: {{ include "sonarqube.agentKeyDerivation.serviceAccountName" . }}
+      {{- with .Values.agentKeyDerivation.securityContext }}
+      securityContext: {{- toYaml . | nindent 8 }}
+      {{- end }}
```
hatem-amairi-sonarsource, line 71: We should expose nodeSelector/tolerations/affinity and sonarqube.agent.scheduling — which every other agentic workload uses and which falls back to the top-level.

On a cluster where the release is pinned to a tainted or labelled node pool, this pre-install hook is unschedulable

hatem-amairi-sonarsource (commented):
Hello @carminevassallo , Nice work!
Added you some comments, most of them applies to the standard chart also.

---

For OpenShift. The hook pod falls to restricted-v2, UID 900 is outside the namespace range, admission rejects it, and the install fails.

charts/sonarqube-dce/templates/agent-key-derivation-hook.yaml
```diff
@@ -0,0 +1,173 @@
...
+              # Upsert: PUT first (the common case is an upgrade over an existing Secret), fall
+              # back to POST when it isn't there yet. Any other status is fatal - notably 403,
+              # which means the Role didn't apply and silently continuing would leave the
+              # consumers mounting nothing.
+              api_write() {
+                curl -sS -o "$RESP" -w '%{http_code}' \
+                  --cacert "$SA/ca.crt" \
+                  -H "Authorization: Bearer $TOKEN" \
+                  -H 'Content-Type: application/json' \
+                  -X "$1" --data "$3" "$2"
```
hatem-amairi-sonarsource, line 119: `--data "$3"` puts every derived key into `/proc/<pid>/cmdline`, readable by anything with host /proc access.

charts/sonarqube-dce/templates/agent-key-derivation-rbac.yaml
carminevassallo: Good catch, dropped the unused get verb and split the rule so update is now scoped via resourceNames to exactly the Secrets the enabled consumers write; create has to stay namespace-wide since Kubernetes can't scope it by name before the object exists.

charts/sonarqube-dce/templates/_helpers.tpl
carminevassallo: Fixed to truncate the sonarqube.fullname prefix before appending the -agent-key-derivation suffix (same pattern as keySecretName), so it can no longer collapse onto the chart's own fullname once the release name is long.

charts/sonarqube-dce/templates/agent-key-derivation-hook.yaml
carminevassallo: Wired the hook Job into the same sonarqube.agent.scheduling helper the orchestrator/runtimes use, with new agentKeyDerivation.nodeSelector/.tolerations/.affinity values documented in the READMEs.

charts/sonarqube-dce/templates/agent-key-derivation-hook.yaml
carminevassallo: Switched api_write() to drive curl from a config file (curl -K -) instead of -H/--data arguments, so neither the bearer token nor the derived key material ever appears in argv.

charts/sonarqube-dce/templates/agent-key-derivation-hook.yaml
carminevassallo: Agreed this is out of scope for this PR. Added a README note that rotating agenticSigningSecret is a hard cutover today, and I'm happy to file a follow-up ticket if we want real rotation support.

matteo-mara-sonarsource (approved):
LGTM.
Deeply discussed offline to ensure everything is addressed.

PR #939, opened 2026-09-03
EA-872 Implement KEDA based autoscaling in helm chart

isuruperera:
Part of EA-872 Implement KEDA based autoscaling in helm chart

Documented the changes and testing notes here: https://xtranet-sonarsource.atlassian.net/wiki/spaces/Platform/pages/5545754634/Agentic+Runtime+-+AutoScaling


charts/sonarqube-dce/values.yaml
```diff
@@ -1092,6 +1092,23 @@ agentOrchestrator:
...
+    metrics:
+      - type: Resource
+        resource:
+          name: cpu
+          target:
+            type: Utilization
+            averageUtilization: 75
+    behavior:
+      scaleDown:
+        stabilizationWindowSeconds: 300
```
carminevassallo, line 1128: how did you calibrate those values?

charts/sonarqube-dce/values.yaml
```diff
@@ -1253,6 +1284,19 @@ remediationAgent:
...
+    enabled: false
+    minReplicas: 2
+    maxReplicas: 25
+    manageReplicas: true
+    pollingInterval: 15
+    scaleDownStabilizationSeconds: 600
+  # Remediation jobs run up to ~30min; must stay comfortably above remediation-agent-runtime's
+  # own fixed `uvicorn --timeout-graceful-shutdown 3600`s (its Dockerfile), or Kubernetes SIGKILLs
+  # the pod before that bounded wait finishes. Don't set SHUTDOWN_GRACE_SECONDS - the image sizes it itself.
+  terminationGracePeriodSeconds: 3630
```
carminevassallo, line 1317: Is the SIGTERM signal handled in the application code? This would protect the running jobs despite k8s sending the termination request...

charts/sonarqube-dce/values.yaml
```diff
@@ -1468,3 +1512,10 @@ gvisor:
 
 # (DEPRECATED) This value is not used in the templates.
 terminationGracePeriodSeconds: 60
+
+# Cluster-capability guard for the KEDA-based agent runtime ScaledObjects (hunterAgent/
+# remediationAgent.autoscaling above).
+agentKeda:
```
carminevassallo, line 1606: I think we should be consistent with what we do with gVisor. We should provide a way to install KEDA from the chart. That's why I'd like to provide KEDA as optional dependency. We recommend people to install it by themselves (this should be the production recommendation) but if they like they can conveniently install it from the chart, wdyt?

E.g.,

```
dependencies:
  - name: keda
    version: xxx
    repository: xxx
    condition: keda.enabled
```

carminevassallo (commented):
Thanks for the PR, @isuruperera!

A few initial comments before testing it on a cluster.

charts/sonarqube-dce/values.yaml
isuruperera: Hello!
These values were based on the testing I did. Documented [here](https://xtranet-sonarsource.atlassian.net/wiki/spaces/Platform/pages/5545754634/Agentic+Runtime+-+AutoScaling)

Orchestrator CPU Utilization: 75% looked to be ideal. Having anything below that seemed aggressively scaling, based on testing, 75% target tracking seemed to scale the orchestrator well for both scheduler and backlog jobs (look at the resource utilization charts)

5min stabilization window was selected because of bursting behavior scheduler.

charts/sonarqube-dce/values.yaml
isuruperera: YES! Agentic runtimes now have a grace period to finish the ongoing jobs when SIGTERM is issued. Also, they will stop accepting new jobs once marked for termination. It's already merged to [master](https://github.com/SonarSource/sonarqube-unification/blob/5bf37f566246e379a6706ebd97903e8daa3b7c76/agentic-workflows/private/agent-runtime/agent_runtime/main.py#L60).

charts/sonarqube-dce/values.yaml
isuruperera: I thought about this approach, but noticed it's not recommended in multiple places due to following risks.

1. KEDA is a cluster wide resource, so when installing the same chart in 2 namespaces, 2nd installation will fail. 
2. When uninstalling our helm chart, it will take down all the `scaledobjects` in the cluster, if customer uses KEDA for some other reason in their cluster. This may silently cripple autoscaling for all other resources who started using keda after we installed it from our chart.

I'm happy to discuss or add it as a chart dependency if we think 2 reasons above are acceptable?

Note about gVisor installer, although gVisor is a cluster wide resource, our installation script can identify if gVisor is present in the node, and safely ignore the re-installation without failing. So, this will not have the 2 risks I mentioned above.

carminevassallo: Hi @isuruperera,

was discussing with PMs today. We think it would be best to have KEDA as an optional dependency. Basically, our chart should provide a way to install the dependency based on a feature flag. In this way, we ease the job of users willing to set up the autoscaling (also those that want to evaluate the feature before purchasing the pack).

Still, we will recommend to install the feature by themselves for advanced users. This will avoid the issues you mentioned as we expect more mature users to operate multiple apps within the same cluster.

charts/sonarqube-dce/values.yaml
carminevassallo: Awesome!

charts/sonarqube-dce/values.yaml
isuruperera: Thanks for helping to clarify this point.
I added KEDA as an optional chart dependency, and enabled it in our Kind tests.
However, didn't enable it for OpenShift tests because the ServiceAccount doesn't allow installing it in the tests.

carminevassallo (approved):
Thanks for addressing the remaining comments. I was testing locally, the autoscaling gets setup as expected.

PR #945, opened 2026-09-07
SONAR-31262 Guard DCE Helm upgrades across ES majors

hatem-amairi-sonarsource:
## Summary
- DCE `helm upgrade` across an Elasticsearch major (2026.1 LTA → 2026.5 LTA) fails while search pods are still running.
- Scale `searchNodes.replicaCount` to 0 on the current chart, then upgrade. Indexes rebuild into `es9`; in-place rollback is not supported.

https://sonarsource.atlassian.net/browse/SONAR-31262

## Test plan
- [x] Unit tests (`TestSearchEsMajor`)
- [x] Kind e2e (`SONARQUBE_ES_MAJOR_E2E=1`): 2026.1.5 install, naive upgrade fails, scale-to-0, 2026.5 upgrade, same-major upgrade still rolls

### DIFF
--- a/charts/sonarqube-dce/README.md
+++ b/charts/sonarqube-dce/README.md
@@ -85,7 +85,12 @@ A rolling update of the search StatefulSet cannot cross an Elasticsearch major v
 helm upgrade -n sonarqube-dce <yourReleaseName> <yourCurrentChart> --reuse-values --set searchNodes.replicaCount=0
 ```
 
-2. Upgrade to the new chart (default `searchNodes.replicaCount` is 3). Do not change the image tag in step 1.
+2. Upgrade to the new chart and restore the search replica count. Do not change the image tag in step 1:
+
+```bash
+helm upgrade -n sonarqube-dce <yourReleaseName> <yourNewChart> --reuse-values --set searchNodes.replicaCount=3
+```
+
 3. After search is Ready, browse to `/setup` and follow the instructions. Indexes are rebuilt into `es9`; rolling back in place is not supported.
 
 The upgrade fails while search pods are still running on the previous Elasticsearch major.
--- a/charts/sonarqube-dce/README.md
+++ b/charts/sonarqube-dce/README.md
@@ -75,6 +75,21 @@ When upgrading to the 2025.6 LTA version, you will experience a few changes.
 4. Browse to <http://yourSonarQubeServerURL/setup> and follow the setup instructions
 5. Reanalyze your projects to get fresh data
 
+### Elasticsearch 8 to 9 (2026.1 LTA to 2026.5 LTA)
+
+A rolling update of the search StatefulSet cannot cross an Elasticsearch major version. Plan a maintenance window and back up the database first.
+
+1. Scale search to 0 with your **current** chart and wait until the search pods are gone:
+
+```bash
+helm upgrade -n sonarqube-dce <yourReleaseName> <yourCurrentChart> --reuse-values --set searchNodes.replicaCount=0
+```
+
+2. Upgrade to the new chart (default `searchNodes.replicaCount` is 3). Do not change the image tag in step 1.
+3. After search is Ready, browse to `/setup` and follow the instructions. Indexes are rebuilt into `es9`; rolling back in place is not supported.
+
+The upgrade fails while search pods are still running on the previous Elasticsearch major.
+
 ### Upgrade from versions prior to 2026.1.0
 
 > **Note**: If you are not using the PostgreSQL dependency (`postgresql.enabled=false`), you can skip this section.
--- a/charts/sonarqube-dce/templates/_helpers.tpl
+++ b/charts/sonarqube-dce/templates/_helpers.tpl
@@ -1389,10 +1389,7 @@ Fail helm upgrade across an Elasticsearch major while search pods are still runn
 {{- $sts := lookup "apps/v1" "StatefulSet" .Release.Namespace $stsName -}}
 {{- if $sts -}}
 {{- $currentMajor := "" -}}
-{{- with $sts.metadata.annotations -}}
-{{- $currentMajor = index . "sonarqube.datacenter/elasticsearch-major" | default "" | toString -}}
-{{- end -}}
-{{- if and (not $currentMajor) $sts.spec $sts.spec.template $sts.spec.template.spec $sts.spec.template.spec.containers -}}
+{{- if and $sts.spec $sts.spec.template $sts.spec.template.spec $sts.spec.template.spec.containers -}}
 {{- $searchContainerName := printf "%s-search" .Chart.Name -}}
 {{- range $sts.spec.template.spec.containers -}}
 {{- if eq (.name | toString) $searchContainerName -}}
--- a/charts/sonarqube-dce/templates/_helpers.tpl
+++ b/charts/sonarqube-dce/templates/_helpers.tpl
@@ -1356,3 +1356,64 @@ periodSeconds: {{ . }}
 timeoutSeconds: {{ . }}
 {{- end }}
 {{- end -}}
+
+{{/*
+Elasticsearch major (8 or 9) inferred from a search image tag.
+*/}}
+{{- define "sonarqube.search.esMajorFromTag" -}}
+{{- $tag := . | toString -}}
+{{- $ver := regexFind "^[0-9]+\\.[0-9]+" $tag -}}
+{{- if $ver -}}
+{{- $year := int (split "." $ver)._0 -}}
+{{- $minor := int (split "." $ver)._1 -}}
+{{- if or (gt $year 2026) (and (eq $year 2026) (ge $minor 4)) -}}
+9
+{{- else -}}
+8
+{{- end -}}
+{{- end -}}
+{{- end -}}
+
+{{- define "sonarqube.search.esMajor" -}}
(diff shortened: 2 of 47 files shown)


charts/sonarqube-dce/README.md
```diff
@@ -75,6 +75,21 @@ When upgrading to the 2025.6 LTA version, you will experience a few changes.
...
+1. Scale search to 0 with your **current** chart and wait until the search pods are gone:
+
+```bash
+helm upgrade -n sonarqube-dce <yourReleaseName> <yourCurrentChart> --reuse-values --set searchNodes.replicaCount=0
+```
+
+2. Upgrade to the new chart (default `searchNodes.replicaCount` is 3). Do not change the image tag in step 1.
+3. After search is Ready, browse to `/setup` and follow the instructions. Indexes are rebuilt into `es9`; rolling back in place is not supported.
+
+The upgrade fails while search pods are still running on the previous Elasticsearch major.
```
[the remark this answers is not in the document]
hatem-amairi-sonarsource: Intentional. Search images from 2026.4.0 run ES 9, so the detector uses that cut. We are not patching the 2026.4 chart (not an LTA). The supported path this chart ships is 2026.1 LTA → 2026.5 LTA, which is what the README describes. 2026.4 → 2026.5 is same ES major, so the guard does not fire. 2026.1 → 2026.4 has no guard because that chart is already out.

charts/sonarqube-dce/values.yaml
```diff
@@ -57,6 +57,8 @@ searchNodes:
 
   replicaCount: 3
 
+  skipEsMajorUpgradeCheck: false
```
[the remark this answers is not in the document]
hatem-amairi-sonarsource: Intentional. This is an unsafe override to bypass the guard, not something we want operators to discover in the values table. Default stays false. If someone actually needs it they can set searchNodes.skipEsMajorUpgradeCheck=true.

tests/unit-test/es_major_upgrade_test.go
```diff
@@ -0,0 +1,68 @@
...
+		})
+	}
+}
+
+func TestSearchEsMajorUpgradeCheckSkippedOnInstall(t *testing.T) {
+	_, err := renderDCESearchWithValues(t, map[string]string{
+		"searchNodes.image.tag": "2026.1.5-datacenter-search",
+	})
+	require.NoError(t, err)
+}
```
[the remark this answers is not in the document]
hatem-amairi-sonarsource: helm template always has IsUpgrade=false and lookup empty, so the fail branch cannot be exercised in the unit/fixture jobs. Annotation inference is covered by TestSearchEsMajor. The real upgrade path is TestDCEEsMajorUpgrade behind SONARQUBE_ES_MAJOR_E2E=1 (needs a kind cluster sized for 3 search nodes); it was run locally (naive upgrade fails, scale-to-0, 2026.5 succeeds). We are not adding that to the default CI workflow.

charts/sonarqube-dce/templates/_helpers.tpl
```diff
@@ -1356,3 +1356,64 @@ periodSeconds: {{ . }}
...
+{{- include "sonarqube.search.esMajorFromTag" (tpl (.Values.searchNodes.image.tag | toString) .) -}}
+{{- end -}}
+
+{{/*
+Fail helm upgrade across an Elasticsearch major while search pods are still running.
+*/}}
+{{- define "sonarqube.search.assertEsMajorUpgrade" -}}
+{{- if and .Release.IsUpgrade (not .Values.searchNodes.skipEsMajorUpgradeCheck) -}}
+{{- $stsName := printf "%s-search" (include "sonarqube.fullname" .) -}}
+{{- $sts := lookup "apps/v1" "StatefulSet" .Release.Namespace $stsName -}}
```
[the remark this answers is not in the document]
hatem-amairi-sonarsource: Helm upgrade already GETs the search StatefulSet for the 3-way merge, so that permission is not new. list pods is only used after an ES major change, to catch a scale-to-0 that has not finished. The kubeconfig that manages this release can list pods in that namespace in the setups we support. Helm 4 lookup does surface Forbidden as a template error; skipEsMajorUpgradeCheck is the override for the rare installer that cannot list pods. We are not documenting that in the public README table.

charts/sonarqube-dce/templates/_helpers.tpl
```diff
@@ -1356,3 +1356,67 @@ periodSeconds: {{ . }}
...
+{{- if and $sts.spec $sts.spec.template $sts.spec.template.spec $sts.spec.template.spec.containers -}}
+{{- $searchContainerName := printf "%s-search" .Chart.Name -}}
+{{- range $sts.spec.template.spec.containers -}}
+{{- if eq (.name | toString) $searchContainerName -}}
+{{- $currentMajor = include "sonarqube.search.esMajorFromTag" (regexFind "[^:]+$" (.image | toString)) -}}
+{{- end -}}
+{{- end -}}
+{{- end -}}
+{{- $targetMajor := include "sonarqube.search.esMajor" . -}}
+{{- if and $currentMajor $targetMajor (ne $currentMajor $targetMajor) -}}
```
jimil09, line 1404: If the regex can't parse the current or target image tag, esMajorFromTag returns "", and this and treats that the same as "nothing to check" — so the upgrade goes through unblocked. We shouldn't skip the check just because we failed to read the tag; we
  should fail instead, since we can't tell if it's safe.

  Example: current STS is on tag 9.9.0-datacenter-search. The regex ^[0-9]+\.[0-9]+ matches "9.9", giving $year = 9. Since 9 >= 2025 is false, esMajorFromTag returns "" (no else branch for that check) instead of erroring. $currentMajor ends up "", so and 
  $currentMajor $targetMajor (...) is false before it even compares majors — the guard silently skips, and the upgrade proceeds as a normal rolling update with no scale-down, even if the actual ES major changed underneath it.

  suggestion:
  `{{- if or (not $currentMajor) (not $targetMajor) (ne $currentMajor $targetMajor) -}}`

charts/sonarqube-dce/templates/_helpers.tpl
```diff
@@ -1356,3 +1356,67 @@ periodSeconds: {{ . }}
...
+{{- include "sonarqube.search.esMajorFromTag" (tpl (.Values.searchNodes.image.tag | toString) .) -}}
+{{- end -}}
+
+{{/*
+Fail helm upgrade across an Elasticsearch major while search pods are still running.
+*/}}
+{{- define "sonarqube.search.assertEsMajorUpgrade" -}}
+{{- if and .Release.IsUpgrade (not .Values.searchNodes.skipEsMajorUpgradeCheck) -}}
+{{- $stsName := printf "%s-search" (include "sonarqube.fullname" .) -}}
+{{- $sts := lookup "apps/v1" "StatefulSet" .Release.Namespace $stsName -}}
```
jimil09, line 1391: Helm's lookup returns empty on RBAC denial, not an error — so "StatefulSet doesn't exist (fresh install)" and "StatefulSet exists but this identity can't get/list it" look identical here. If the principal running helm upgrade lacks RBAC on
  apps/v1/statefulsets (or, separately, on v1/pods at line 1407), $sts (or $podList) comes back nil and the whole guard — including the replicas > 0 check — is silently skipped during a real cross-major upgrade.

  Example: a GitOps operator (Argo CD/Flux) or CI pipeline deploys this chart with a ServiceAccount scoped only to the resources this chart's own templates/ creates. It's plausible that ServiceAccount has no read access to StatefulSets/Pods cluster- or
  namespace-wide beyond what it manages. On such a setup: lookup "apps/v1" "StatefulSet" ... returns nil (not an error) → if $sts is false → the entire guard body, including the fail, never executes → an ES8→ES9 upgrade proceeds with live search pods
  still up, no warning printed anywhere.

charts/sonarqube-dce/templates/_helpers.tpl
hatem-amairi-sonarsource: Hello, this is working as intended, but the intent isn't explicit in the code, so it reads like an oversight. We deliberately don't run this check on unsupported tags (custom tags, or versions from the old pre-calendar versioning scheme), since we only support LTA → LTA upgrade paths and can't reliably infer the Elasticsearch version from the chart version once the tag falls outside that scheme. 
I'll make that intent explicit in the code!

charts/sonarqube-dce/templates/_helpers.tpl
hatem-amairi-sonarsource: Thanks for flagging! Checked against Helm's `lookup` source at our pinned `v4.0.5`: only `NotFound` resolves to empty and `Forbidden` returns as `err` and hard-fails the render, it doesn't fall through silently. https://github.com/helm/helm/blob/v4.0.5/pkg/engine/lookup_func.go#L71-L86

Gitar-bot hit this same path earlier with the mechanism the other way round (Forbidden fails the upgrade, doesn't skip the guard), which matches: https://github.com/SonarSource/helm-chart-sonarqube/pull/945#discussion_r3952229859

If you've got a repro showing the silent-skip, happy to dig further :)

charts/sonarqube-dce/templates/_helpers.tpl
jimil09: Fair point.

jimil09 (approved):
LGTM

PR #949, opened 2026-09-09
SONAR-32198 Setup Kubernetes Horizontal Pod Autoscaling (HPA) for Agentic Analysis

isuruperera:
Part of SONAR-32198 Setup Kubernetes Horizontal Pod Autoscaling (HPA) for Agentic Analysis

Notes on Testing
Triggered a large number of analysis requests to test the autoscaling.

Pods were autoscaled up to the max configured, and scaled down to 3 pods once the load was finished.
image

image


leo-geoffroy-sonarsource:
**Heads-up: the autoscaling algorithm differs between Server (this chart) and Cloud.**

Server's `ScaledObject` (`vortex-scaledobject.yaml`) sums each replica's `max-concurrent-requests` peak across the fleet (`aggregateFromKubeServiceEndpoints` + `aggregationType: sum`), then divides by `targetConcurrentRequests`:

```
desiredReplicas = ceil(sum(maxConcurrentRequests) / targetConcurrentRequests)
```

Cloud's autoscaling instead **averages** `maxConcurrentRequests` across instances rather than summing.
Flagging this, even though I'm not sure if this as an impact

```
desiredReplicas = ceil(sum(maxConcurrentRequests) / targetConcurrentRequests)
```
should be equivalent to 
```
currentConcurrentRequests = sum(maxConcurrentRequests) / currentReplicas
```
with `targetConcurrentRequests` tends to `currentConcurrentRequests`

**Side note:** Cloud also allows `targetConcurrentRequests` to be a decimal (e.g. `1.5`). Here, `values.schema.json` types it as `integer`, so a fractional value is rejected by Helm's schema validation before it ever reaches KEDA — another point of divergence between the two, in case decimal precision turns out to matter for tuning.

leo-geoffroy-sonarsource:
> **Heads-up: the autoscaling algorithm differs between Server (this chart) and Cloud.**
> 
> Server's `ScaledObject` (`vortex-scaledobject.yaml`) sums each replica's `max-concurrent-requests` peak across the fleet (`aggregateFromKubeServiceEndpoints` + `aggregationType: sum`), then divides by `targetConcurrentRequests`:
> 
> ```
> desiredReplicas = ceil(sum(maxConcurrentRequests) / targetConcurrentRequests)
> ```
> 
> Cloud's autoscaling instead **averages** `maxConcurrentRequests` across instances rather than summing. Flagging this, even though I'm not sure if this as an impact
> 
> ```
> desiredReplicas = ceil(sum(maxConcurrentRequests) / targetConcurrentRequests)
> ```
> 
> should be equivalent to
> 
> ```
> currentConcurrentRequests = sum(maxConcurrentRequests) / currentReplicas
> ```
> 
> with `targetConcurrentRequests` tends to `currentConcurrentRequests`
> 
> **Side note:** Cloud also allows `targetConcurrentRequests` to be a decimal (e.g. `1.5`). Here, `values.schema.json` types it as `integer`, so a fractional value is rejected by Helm's schema validation before it ever reaches KEDA — another point of divergence between the two, in case decimal precision turns out to matter for tuning.

Confirmed by @isuruperera  and @carminevassallo , it should be equivalent in term of behavior

charts/sonarqube/templates/validation.yaml
```diff
@@ -165,6 +167,46 @@ It is used to validate the values.yaml file before the installation starts.
...
 {{- fail "\n ** The values.yaml file is not valid. ** \n vortex.enabled is true but only one of vortex.storage.accessKey / vortex.storage.secretKey is set. Provide both, or neither for credential-less access." -}}
   {{- end -}}
+  {{- $vortexAutoscaling := .Values.vortex.autoscaling -}}
+  {{- if $vortexAutoscaling.enabled -}}
+    {{- /* Without cross-replica aggregation, KEDA samples a single random pod through the
+           Service, so the sum-vs-target arithmetic is only valid for exactly one replica - the
+           normal >= 2 floor would make that fallback impossible to configure at all. */}}
+    {{- $vortexMinReplicaFloor := 2 -}}
+    {{- if not $vortexAutoscaling.aggregateAcrossReplicas -}}
+      {{- $vortexMinReplicaFloor = 1 -}}
```
jacek-poreda-sonarsource: The floor relaxes to 1 but nothing caps `minReplicas`/`maxReplicas`. `aggregateAcrossReplicas: false` with the shipped `minReplicas: 3`/`maxReplicas: 20` renders fine — that's the config the README calls "only correct with `minReplicas: 1`", i.e. a 3–20 pod fleet scaled off one random pod's peak, silently.

Either fail here when `minReplicas != 1`, or soften the README: the fallback is a 1/N estimate at any N > 1, so `minReplicas: 1` only guarantees the *starting* sample is accurate. Same in `sonarqube-dce`.

charts/sonarqube/values.schema.json
```diff
@@ -277,6 +277,50 @@
...
+                        },
+                        "targetConcurrentRequests": {
+                            "type": "integer",
+                            "minimum": 1
+                        },
+                        "aggregateAcrossReplicas": {
+                            "type": "boolean"
+                        },
+                        "windowSeconds": {
+                            "type": ["integer", "string"]
```
jacek-poreda-sonarsource: `windowSeconds: 0` is falsy in Go templates, so it silently skips both the `pollingInterval <= windowSeconds` guard and the env var. Add `"minimum": 1` like sibling `pollingInterval`. Same in `sonarqube-dce`.

charts/sonarqube/templates/vortex-scaledobject.yaml
```diff
@@ -0,0 +1,48 @@
...
+  pollingInterval: {{ .Values.vortex.autoscaling.pollingInterval }}
+  advanced:
+    horizontalPodAutoscalerConfig:
+      {{- /* KEDA otherwise derives the HPA name as "keda-hpa-<scaledobject name>", whose 9-char
+             prefix can push it past the 63-char name limit even though the ScaledObject name
+             itself fits. Reuse the (already-truncated) ScaledObject name directly instead. */}}
+      name: {{ include "sonarqube.vortex.fullname" . }}
+      behavior:
+        scaleDown:
+          # Cloud's own scale-in cooldown, and the guard against KEDA's partial-endpoint-failure
```
jacek-poreda-sonarsource: `#` instead of `{{- /* */}}`, so this lands in the rendered manifest — internal "Cloud's own" reference included. Every other comment in this file is template-side.

charts/sonarqube/templates/vortex-scaledobject.yaml
```diff
@@ -0,0 +1,48 @@
...
+           design. */}}
+    - type: metrics-api
+      metadata:
+        {{- /* FQDN, not sonarqube.vortex.url's short name: KEDA parses this host as
+               <service>.<namespace> to find the Service's EndpointSlices, and rejects a bare short
+               name outright. The port must be the *pod* port, not just any port on the Service -
+               here they're always the same number (.Values.vortex.port), so this is always safe. */}}
+        url: "http://{{ include "sonarqube.svcFQDN" (dict "name" (include "sonarqube.vortex.fullname" .) "ctx" .) }}:{{ .Values.vortex.port }}{{ .Values.vortex.autoscaling.metricPath }}"
+        valueLocation: "value"
+        targetValue: "{{ .Values.vortex.autoscaling.targetConcurrentRequests }}"
```
jacek-poreda-sonarsource, line 55: Worth pinning `metricType: AverageValue` explicitly. The whole per-replica-target arithmetic rests on it, and the KEDA metrics-api docs don't document a default `metricType` at all.

charts/sonarqube/templates/vortex-scaledobject.yaml
```diff
@@ -0,0 +1,48 @@
+{{- if and .Values.vortex.enabled .Values.vortex.autoscaling.enabled }}
+apiVersion: keda.sh/v1alpha1
+kind: ScaledObject
+metadata:
+  name: {{ include "sonarqube.vortex.fullname" . }}
+  labels: {{- include "sonarqube.labels" . | nindent 4 }}
+    app.kubernetes.io/component: vortex
+spec:
```
jacek-poreda-sonarsource, line 8: No `spec.fallback`. The 900s window absorbs *partial* poll failures, but a total scrape failure leaves the HPA with no metric — worth making that path explicit.

charts/sonarqube/values.yaml
```diff
@@ -770,10 +770,49 @@ vortex:
...
     type: Recreate
 
+  # Peak-concurrency autoscaling via KEDA, reading Vortex's own unauthenticated
+  # GET /metrics/max-concurrent-requests. Requires the KEDA CRDs in-cluster - see agentKeda below.
+  autoscaling:
+    enabled: false
+    # Validated >= 2 (see validation.yaml), or >= 1 with aggregateAcrossReplicas: false below:
+    # this fleet never scales to zero, so there is no 0<->1 transition to reason about. Note this
+    # raises a default replicaCount: 1 install to 3 pods.
+    minReplicas: 3
```
jacek-poreda-sonarsource, line 785: Why 3? Every sibling defaults to `minReplicas: 2`, and this is the one value in the block without a rationale comment.

tests/unit-test/test-cases-values/sonarqube/vortex-autoscaling.yaml
```diff
@@ -0,0 +1,41 @@
+community:
+  enabled: true
+monitoringPasscode: test-passcode
+image:
+  pullSecret: app-node-secret
+  pullSecrets:
+    - name: app-node-secret-list
+networkPolicy:
+  enabled: true
+# Vortex enabled with autoscaling on top of the vortex-enabled.yaml fixture: RollingUpdate is
```
jacek-poreda-sonarsource, line 10: Not actually layered on `vortex-enabled.yaml` — it's a standalone copy, and it drops that fixture's `nodeSelector`/`affinity`/`tolerations`.

jacek-poreda-sonarsource (approved):
Approving — KEDA usage checks out against upstream 2.20 docs (FQDN + pod-port requirement, `aggregationType: sum` + AverageValue arithmetic), the bundled subchart is 2.20.2 so it already clears the documented floor, both charts are byte-identical, and test coverage is thorough. Comments below are mostly small; the `aggregateAcrossReplicas` one is worth closing before or right after merge.

One doc suggestion not tied to a line: the README never connects the bundled keda 2.20.2 to the `>= 2.20.0` requirement — saying so tells `keda.enabled: true` users they never need the fallback at all.

charts/sonarqube/values.yaml
isuruperera: Got it as a requirement from Vortex: https://sonarsource.atlassian.net/browse/SONAR-32198

charts/sonarqube/templates/validation.yaml
isuruperera: Thanks! Updated the validation to fail in these edge cases!

charts/sonarqube/values.schema.json
isuruperera: Added "minimum": 1

charts/sonarqube/templates/vortex-scaledobject.yaml
isuruperera: Added the fallback path. If total scrape fails, HPA sets to current replicas instead of degrading

charts/sonarqube/templates/vortex-scaledobject.yaml
isuruperera: Thanks! I've set the value to AverageValue explicitly

charts/sonarqube/templates/vortex-scaledobject.yaml
isuruperera: Fixed!

PR #953, opened 2026-09-09
SONAR-32205 Point agent probes at /livez and /readyz endpoints

carminevassallo:
## Summary
- Switch `agentOrchestrator`/`hunterAgent`/`remediationAgent` liveness probes from the aggregate `/health` endpoint to the dedicated `/livez` endpoint (readiness already used `/readyz` for the runtimes; orchestrator now uses both dedicated endpoints too), so a db/storage outage that readiness alone should absorb no longer false-trips liveness and kills the pod.
- Add `agentOrchestrator.terminationGracePeriodSeconds` (default `1320`s), sized against the orchestrator's `spring.lifecycle.timeout-per-shutdown-phase` (`1260`s) plus margin.
- Fix `hunterAgent.terminationGracePeriodSeconds` drift: `14430` → `44430`, so it actually exceeds the image's `44400`s uvicorn graceful-shutdown timeout (it previously undershot it).

## Test plan
- [x] `go test ./...` in `tests/unit-test` — full suite passes
- [x] New `TestAgentOrchestratorProbesCustomized` test + `agent-probes-custom.yaml` fixture (both charts) proving probes are overridable end-to-end, including that `liveness.enabled: false` drops the probe block entirely
- [x] Updated existing probe/grace-period tests (`agent_orchestrator_test.go`, `agent_runtime_test.go`, `agent_autoscaling_test.go`)
- [x] Regenerated golden fixtures — diff limited to the intended probe-path/grace-period lines
- [x] `helm template` smoke checks confirming correct paths/grace-periods on both charts, plus the readiness-path/liveness-disabled override case
- [x] `.github/scripts/unit_helm_compatibility_test.sh` for both charts — all fixtures render valid manifests per `kubeconform`
- [x] `helm lint` clean on both charts
- [x] `README.md`/`CHANGELOG.md` updated in both charts (no `Chart.yaml` version bump)

Jira: https://sonarsource.atlassian.net/browse/SONAR-32205


charts/sonarqube-dce/README.md
```diff
@@ -999,7 +999,9 @@ Four independently toggled features: Vortex (`vortex.enabled`), the shared Agent
...
+- `agentOrchestrator.terminationGracePeriodSeconds` (default `1320`) similarly ships unconditionally. It's sized against the orchestrator image's own `spring.lifecycle.timeout-per-shutdown-phase: 21m` (`1260`s) plus margin — the drain is additive across its two same-phase db-schedulers, not a `max()`.
+- `agentOrchestrator`/`hunterAgent`/`remediationAgent` liveness probes target the dedicated `/livez` endpoint, never the aggregate `/health` — the latter also reports the pod's db/storage connectivity and would falsely trip liveness (and get the pod killed) during an outage that readiness alone should absorb. `/livez`/`/readyz` only exist on images built after EA-937/EA-938; pointing an older image at this chart version 404s on both probes and the pod CrashLoopBackOffs — `image.tag` is operator-supplied and empty by default, so the chart can't enforce a minimum version itself.
```
manuel-vianello-sonarsource: Readiness 404 only holds the pod NotReady; only liveness 404 restarts it.

Same for `charts/sonarqube/README.md:905`.

charts/sonarqube/values.schema.json
```diff
@@ -562,6 +562,10 @@
                             "type": "object"
                         }
                     }
+                },
+                "terminationGracePeriodSeconds": {
```
manuel-vianello-sonarsource, line 566: Nit

minimum: 0 permits a value that would SIGKILL the drain immediately — the exact failure the field exists to prevent. 

Matches the existing runtime schema, so consistent, but minimum: 1 would better encode intent. Same on `charts/sonarqube-dce/values.schema.json`

manuel-vianello-sonarsource (approved):
LGTM ✅ 

Two non-blocking comments.

PR #955, opened 2026-09-09
SONAR-24525 Size chart memory defaults for the new Web/CE heap defaults

lukasz-jarocki-sonarsource:
## Summary
`sonar-enterprise#18818` (SONAR-24525) raises SonarQube's default `-Xmx` for the Web and Compute Engine processes to reduce the probability of hitting an OOM that permanently breaks the bundled Elasticsearch HTTP client (see USER-96):
- Community / Developer: web 512m -> 1G, ce 512m -> 1.5G (out of scope for this chart, see below)
- **Enterprise / Data Center: web 1G -> 2G, ce 2G -> 4G** (this is what `charts/sonarqube` and `charts/sonarqube-dce` are sized for)

This chart already documents a sizing methodology tying `resources.requests/limits.memory` to a per-edition "Sum of Xmx" table and a "~80% rule" (Xmx sum should be ~80% of the pod's memory). Updated accordingly:

- **`charts/sonarqube`** (community/developer/enterprise, single container running web+ce+search): enterprise-tier Xmx sum goes from `1G+2G+2G=5G` to `2G+4G+2G=8G`. Applying the chart's existing ~80% rule: `requests.memory` 2048M -> 4096M, `limits.memory` 6144M -> 10240M (8192M / 10240M = exactly 80%). Updated the doc comment and README's Xmx-sum table/parameter table to match.
- **`charts/sonarqube-dce`** `applicationNodes` (web+ce only, no search on app nodes): Xmx sum goes from `1G+2G=3G` to `2G+4G=6G`. Same request=limit pattern as before: 4096M -> 8192M (6144M/8192M = 75%, matching the existing ratio used for this block). `searchNodes` is untouched since Search's `-Xmx` (2G) did not change.
- Regenerated the `tests/unit-compatibility-test/fixtures` golden snapshots via `.github/scripts/generate_helm_fixtures.sh`. Note: a plain regeneration on current `master` (without any of my changes) already produces unrelated churn in these fixtures (stale `checksum/config` etc., likely a pre-existing helm-version/tooling drift, not something this PR touches) — I diffed a "regenerated from unmodified values" pass against a "regenerated from my edited values" pass and hand-applied only the isolated `resources.*.memory` line changes onto the committed fixtures, so this PR's fixture diff is scoped to exactly the memory values (144 lines across 68 files, 2 lines each) and doesn't carry unrelated pre-existing drift.

## Test plan
- [x] `helm lint charts/sonarqube --set monitoringPasscode=test,edition=enterprise` passes
- [x] `helm lint charts/sonarqube-dce` passes (pre-existing mandatory-value guard messages are unrelated to this change, verified present on unmodified `master` too)
- [x] Fixture diff reviewed to confirm only `resources.*.memory` values changed, no other template output regressed

### DIFF
--- a/charts/sonarqube-dce/README.md
+++ b/charts/sonarqube-dce/README.md
@@ -388,12 +388,12 @@ Please find here the default SonarQube Xmx parameters to setup the memory reques
 | Edition                             | Sum of Xmx |
 | ----------------------------------- | ---------- |
 | datacenter edition searchNodes      | 2G         |
-| datacenter edition applicationNodes | 3G         |
+| datacenter edition applicationNodes | 6G         |
 
 To comply with the 80% rule mentioned above, we set the following default values:
 
 * searchNodes.resources.memory.request/limit=3072M
-* applicationNodes.resources.memory.request/limit=4096M
+* applicationNodes.resources.memory.request/limit=8192M
 
 Please feel free to adjust those values to your needs. However, given that memory is a “non-compressible” resource, we advise you to set the memory requests and limits to the **same**, making memory a guaranteed resource. This is needed especially for production use cases.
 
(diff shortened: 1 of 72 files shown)


jimil09 (changes requested):
We'd need to pin `ApplicationNodes.resources` in `charts/sonarqube-dce/ci/ci-values.yaml` to CI-appropriate values, which is the pattern `charts/sonarqube/ci/ci-values.yaml` already follows. This is because `charts/sonarqube-dce/ci/ci-values.yaml` doesn't override `ApplicationNodes.resources`, so it inherits the new 8192M request from `values.yaml`. With `ApplicationNodes.replicaCount: 3`, that's ~24.6Gi of memory requests on the single-node kind cluster.

PR #956, opened 2026-09-09
SONAR-32220 Pin agent runtime HPA name to avoid 63-char overflow

carminevassallo:
## Summary
- KEDA derives the underlying HPA's name as `keda-hpa-<scaledobject-name>`; that 9-char prefix can push it past Kubernetes' 63-char name limit even when the ScaledObject's own (already-truncated) name fits, and KEDA's admission webhook rejects the ScaledObject outright.
- Pin `spec.advanced.horizontalPodAutoscalerConfig.name` to the ScaledObject's own fullname in both `sonarqube` and `sonarqube-dce` charts so KEDA reuses it verbatim instead of adding the prefix.
- Reproduced with the `next-prod-linas` DCE release (see SONAR-32220).

## Test plan
- [x] `cd tests/unit-test && go test ./...` passes
- [x] Added `TestAgentRuntimeScaledObjectHPANameStaysWithinLimit` reproducing the real failing release name, asserting the HPA name is ≤63 chars
- [x] Verified via `helm template` against the actual failing values file that `horizontalPodAutoscalerConfig.name` now renders within limits

Fixes SONAR-32220

### DIFF
--- a/charts/sonarqube-dce/templates/agent-runtime-scaledobject.yaml
+++ b/charts/sonarqube-dce/templates/agent-runtime-scaledobject.yaml
@@ -21,6 +21,10 @@ spec:
   pollingInterval: {{ $cfg.autoscaling.pollingInterval }}
   advanced:
     horizontalPodAutoscalerConfig:
+      {{- /* KEDA otherwise derives the HPA name as "keda-hpa-<scaledobject name>", whose 9-char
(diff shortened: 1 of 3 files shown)


navyatejaotturu-sonarsource (approved):
Looks good to me!

PR #962, opened 2026-09-14
SONAR-32320 Merge jvmOpts with sonar.*.javaOpts

hatem-amairi-sonarsource:
`applicationNodes.jvmOpts`/`jvmCeOpts` (sonarqube-dce) and `jvmOpts`/`jvmCeOpts` (sonarqube) were
silently dropped whenever sonarProperties also set `sonar.web.javaOpts/sonar.ce.javaOpts` -
risking an unintended default JVM heap. Now both values are concatenated (space-separated) into
the resulting `SONAR_WEB_JAVAOPTS`/`SONAR_CE_JAVAOPTS` env var instead of one overwriting the other.
Non-breaking: when only one of the two is set, behavior is unchanged. Fixed in both charts since
they share the same helper logic.

## Test plan
- [x] `go test ./...` in tests/unit-test (new jvm_opts_test.go, both charts)
- [x] kubeconform compatibility gate passes for both charts
- [x] `helm install --dry-run=server` against a local kind cluster confirms the merged
      SONAR_WEB_JAVAOPTS/SONAR_CE_JAVAOPTS


navyatejaotturu-sonarsource (approved):
Looks good to me. 👍

PR #966, opened 2026-09-15
SONAR-32379 Default agentic Deployments to RollingUpdate

carminevassallo:
## Summary
- Vortex, orchestrator, hunter, and remediation Deployments now default to `strategy.type: RollingUpdate` instead of `Recreate` (Vortex) or a blank strategy (the other three), so upgrades no longer take the whole agentic fleet down at once.
- `Recreate` remains available as an explicit override; `validation.yaml` still rejects it whenever `vortexAnalysis.autoscaling.enabled=true`.
- Regenerated affected golden fixtures, updated Go unit tests, and updated the `strategy` table rows in both charts' READMEs.

Jira: https://sonarsource.atlassian.net/browse/SONAR-32379

## Test plan
- [x] `go build ./...`
- [x] `go test ./...`
- [x] `helm lint charts/sonarqube --set community.enabled=true`
- [x] `helm lint charts/sonarqube-dce --set community.enabled=true`
- [x] Regenerated fixtures via `.github/scripts/generate_helm_fixtures.sh` and confirmed only the expected 4 files per chart changed


carminevassallo:
Reopening against feature/unified-agentic-harness from a dedicated branch instead.

PR #969, opened 2026-09-17
SONAR-32431 Allow vortexAnalysis.autoscaling.targetConcurrentRequests to be a float

leo-geoffroy-sonarsource:
## Summary
- `vortexAnalysis.autoscaling.targetConcurrentRequests` was restricted to a whole number (`values.schema.json` `type: integer`, and `validation.yaml` cast it with `int`), which blocked customers from fine-tuning the KEDA scale-out threshold in small increments.
- KEDA's `metrics-api` scaler parses `targetValue` as a float (`strconv.ParseFloat`), so the chart was stricter than the tool it configures.
- Loosens the schema to `type: number` and the validation guard to compare as `float64` (both `sonarqube` and `sonarqube-dce` charts). The `minimum: 1` floor is unchanged.

Fixes SONAR-32431.

## Test plan
- [x] `helm template` with `targetConcurrentRequests: 1.5` passes schema validation and renders `targetValue: "1.5"` on the KEDA `ScaledObject` trigger.
- [x] `helm template` with `targetConcurrentRequests: 0.5` is still correctly rejected by the schema `minimum: 1`.
- [ ] Go unit test suite (`tests/unit-test`) — could not run locally (no `go` toolchain available in this environment); existing `vortex_autoscaling_test.go` cases pass integer-literal strings (e.g. `"12"`) which remain valid under the loosened schema.


carminevassallo (approved):
LGTM!

PR #970, opened 2026-09-17
SONAR-32432 Block Hunter Agent from reaching SonarQube via egress

carminevassallo:
## Summary
- The Agent Egress Proxy (Squid) had a single hardcoded allow rule for SonarQube's `/api/rules/show` and `/api/v2/a3s/private/analyses` endpoints, reachable by any agent runtime — not just Remediation as intended. Hunter was reachable to SonarQube in practice, not just in theory.
- Splits the shared Squid process into two listening ports per family: Hunter keeps `agentEgressProxy.port` (loses the SonarQube allow rule), Remediation gets a new `agentEgressProxy.remediationPort` (keeps everything Hunter has, plus SonarQube).
- Reinforced with defense-in-depth at the NetworkPolicy layer: the proxy's ingress policy and each family's own egress policy are split so Hunter can only dial the Hunter port and Remediation only the Remediation port, independent of Squid's ACLs.
- Applied identically to `charts/sonarqube` and `charts/sonarqube-dce`.
- No values.yaml opt-out, consistent with the chart's existing pattern for security-critical rules. Gated behind `remediationAgent.enabled` (default `false`), so existing Hunter-only deployments render byte-identical apart from an additional NetworkPolicy selector hardening (see test plan).

Tracked as [SONAR-32432](https://sonarsource.atlassian.net/browse/SONAR-32432), relates to [EA-799](https://sonarsource.atlassian.net/browse/EA-799).

## Test plan
- [x] `helm template` both charts with only `hunterAgent.enabled=true`, diffed against the pre-change rendering via `git stash`: only the NetworkPolicy ingress selector tightened (hardening, no longer family-agnostic), a comment update, and the resulting `checksum/config` change.
- [x] `helm template` both charts with `hunterAgent.enabled=true, remediationAgent.enabled=true`: verified rendered `squid.conf` (two `http_port` lines, `myport` ACL, scoped allow rule), the proxy's `NetworkPolicy` (two ingress rules, correct ports/selectors), each runtime's own `NetworkPolicy` egress rule and injected proxy env vars, and the Deployment/Service (two container ports/Service ports).
- [x] Same, with `istio.meshSidecar.enabled=true` (`--kube-version 1.30.0`, sandboxed): verified both `Sidecar` CRs' egress listener ports.
- [x] Confirmed `charts/sonarqube` and `charts/sonarqube-dce` output differs only in the pre-existing SonarQube-pod-selector difference, nothing new introduced by this change.

[SONAR-32432]: https://sonarsource.atlassian.net/browse/SONAR-32432?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


charts/sonarqube-dce/templates/agent-egress-proxy-configmap.yaml
```diff
@@ -84,6 +94,12 @@ data:
     # still denied rather than reopening the whole API surface.
     acl sonarqube_host dstdomain {{ include "sonarqube.svcFQDN" (dict "name" (include "sonarqube.fullname" .) "ctx" .) }}
     acl sonarqube_agentic_endpoints urlpath_regex ^[^?]*/rules/show(\?|$) ^[^?]*/a3s/private/analyses(/|\?|$)
+    {{- if .Values.remediationAgent.enabled }}
+    # Restricts the sonarqube_host allow rule below to the Remediation-only listener - Hunter,
+    # which only ever dials agentEgressProxy.port, can never match this ACL and so can never reach
+    # SonarQube through this proxy, no matter what allowedDomains or extraSquidConf contains.
+    acl remediation_listener myport {{ .Values.agentEgressProxy.remediationPort }}
```
hatem-amairi-sonarsource: The `acl remediation_listener myport {{ .Values.agentEgressProxy.remediationPort }}` scoping is purely port-based — it has no notion of which family made the request. I stood up the actual rendered `squid.conf` on a real Squid in a kind cluster with a plain, identity-less curl client: dialing port 3129 directly got `200` on `/api/rules/show`, same as a real Remediation pod would. The only thing stopping a Hunter pod from doing the same is the NetworkPolicy split this PR adds — and `agentEgressProxy.networkPolicy.enabled` (plus the per-family equivalents) defaults to `false`. So "Hunter is structurally unable to reach SonarQube... independent of Squid's ACLs" only holds once an operator turns NetworkPolicy on. Is that an accepted tradeoff we should just document (like [`README.md:1071`](https://github.com/SonarSource/helm-chart-sonarqube/blob/72805a9e3112f50f2531b796fb5c070d496120ca/charts/sonarqube-dce/README.md#L1071) already does for the mesh-sidecar case), or should `networkPolicy.enabled` default/force true once `remediationAgent.enabled`? 

**Also applies to the standard chart.**

charts/sonarqube-dce/values.yaml
```diff
@@ -1515,6 +1515,11 @@ agentEgressProxy:
     pullSecret: ""
     pullSecrets: []
   port: 3128
+  # Second listening port on the same Squid process, used exclusively by the Remediation runtime -
+  # it's the only family allowed to reach SonarQube's own agentic endpoints (see
+  # agent-egress-proxy-configmap.yaml's sonarqube_agentic_endpoints rule). Hunter only ever dials
+  # agentEgressProxy.port above, which never carries that allow rule.
+  remediationPort: 3129
```
hatem-amairi-sonarsource, line 1522: No validation guards against `remediationPort` colliding with `agentEgressProxy.port` (or `agentOrchestrator.port`/`istio.meshSidecar.meshPort`). There's precedent for exactly this kind of guard at [`_helpers.tpl:854`](https://github.com/SonarSource/helm-chart-sonarqube/blob/72805a9e3112f50f2531b796fb5c070d496120ca/charts/sonarqube-dce/templates/_helpers.tpl#L854) (fails closed on a meshPort collision) — worth a similar `fail` here, since a collision means Squid tries to bind the same port twice and crash-loops. 

**Also applies to the standard chart.**

charts/sonarqube/templates/agent-egress-proxy-configmap.yaml
hatem-amairi-sonarsource: gitar-bot's first review flagged this and its second review lists it under "5 closed" — but the `acl sonarqube_host`, `acl sonarqube_agentic_endpoints`, and `acl Safe_ports port {{ .Values.service.externalPort }}` still render unconditionally in a Hunter-only build (confirmed by rendering it live), and the `Safe_ports` comment is still stale. No security impact, just dead/misleading config — but flag it since the bot's own tracking says this is done. 

Also on lines L95-L96.

**Also applies to the standard chart.**

hatem-amairi-sonarsource (commented):
Hello @carminevassallo, nice work on this!

I set up a kind cluster with a real Squid instance running the actual rendered `squid.conf` to poke at the ACL scoping directly: `403` on Hunter's port for the SonarQube path, `200` on Remediation's — the port split does what it's meant to.

Added a few comments below on top of what gitar already found. Most are minor, but one's more of a question than a nitpick: whether the "Hunter can't reach SonarQube" guarantee holds without `agentEgressProxy.networkPolicy.enabled`, which defaults to `false` — worth a look.

charts/sonarqube-dce/templates/agent-egress-proxy-configmap.yaml
carminevassallo: Network policies should be indeed set to true by default IMO

charts/sonarqube-dce/templates/validation.yaml
```diff
@@ -77,6 +77,13 @@
   {{- if not .Values.vortexAnalysis.enabled -}}
 {{- fail "\n ** The values.yaml file is not valid. ** \n remediationAgent.enabled is true but vortexAnalysis.enabled is not true." -}}
   {{- end -}}
+  {{- /* agentEgressProxy.remediationPort only opens a second http_port directive in
+         agent-egress-proxy-configmap.yaml's squid.conf when remediationAgent.enabled - both
+         listeners live in the same Squid container, so an equal port is a duplicate bind Squid
+         refuses to start with. */ -}}
+  {{- if eq (int .Values.agentEgressProxy.port) (int .Values.agentEgressProxy.remediationPort) -}}
```
hatem-amairi-sonarsource, line 84: The new guard only catches `agentEgressProxy.remediationPort` colliding with
  `agentEgressProxy.port`. I checked the other two ports this value can plausibly collide with by
  rendering both charts with each collision live:

  - Colliding with `agentOrchestrator.port` (default 8080): renders fine, and I confirmed it's
    actually benign — Squid ends up with two distinct `http_port` numbers, so no bind conflict, and
    the orchestrator's port 8080 lives on a separate Pod/Service entirely. Reusing the number is
    confusing (`remediation_listener` incidentally matches "port 8080" for two unrelated reasons) but
    not a functional break. I don't think this one needs a guard.

  - Colliding with `istio.meshSidecar.meshPort` (default 18080, only reachable with
    `istio.meshSidecar.enabled=true`): this one does produce a real conflict. The rendered `Sidecar`
    CR for the Remediation family ends up with both its `ingress[0].port.number` and
    `egress[0].port.number` set to 18080 in the same resource (one bound `0.0.0.0`, one bound
    `127.0.0.1`) — a duplicate port declaration inside one Sidecar. I didn't stand up Istio/Envoy to
    watch it fail live, so I can't say for certain what Envoy does with it, but two listeners
    declaring the same port number in one Sidecar resource is exactly the shape the existing guard at
    `_helpers.tpl:700` (sonarqube) / `_helpers.tpl:854` (sonarqube-dce) already fails closed on for
    the *runtime's own* port — "fails closed if meshPort collides with an enabled runtime's own port".
    `remediationPort` is functionally another runtime-family port for this exact purpose and seems to
    have been missed when that guard was written (it predates this PR). Might be worth extending that
    same check to `remediationPort`, gated the same way the existing one is (only relevant once
    `istio.meshSidecar.enabled`).

  Not blocking — this needs deliberately setting `remediationPort` to a non-default value that also
  happens to equal `meshPort`, so it's an edge case, not something the shipped defaults hit. Flagging
  because there's already a precedent for exactly this kind of guard right next to where you added the
  new one.
  
  Applies also for the standard chart.

hatem-amairi-sonarsource (approved):

charts/sonarqube-dce/templates/validation.yaml
carminevassallo: Addressed

